PR #1, opened 2024-06-26
Format as package

kaitejohnson:
Not doing anything here but setting up infrastructure for the creation of an R package called `wwinference`. The goal will be to in the next week, port of just the modeling components from `cfaforecastrenewalww` and the immediate pre-processing from `wweval`. This PR should just provide the skeleton components for that package.

Next steps: 
- set up CI for testing suite

PR #2, opened 2024-06-28
styler patch

dylanhmorris:


.github/workflows/pre-commit.yaml
```diff
@@ -13,9 +13,9 @@ jobs:
...
       with:
         python-version: "3.12"
-    - uses: actions/cache@v3
+    - uses: actions/cache@v4
       with:
         path: |
-          ~/.cache/pre-commit
           ~/.cache/R/renv
-        key: pre-commit-3|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
+        key: r-precommit-cache|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
```
kaitejohnson, line 20: So was the key thing here that we needed to point to an `r-precommit-cache`? Just trying to understand this at a high level.

.pre-commit-config.yaml
```diff
@@ -35,7 +35,8 @@ repos:
     rev: v0.4.2
     hooks:
     -   id: style-files
-        args: [--style_pkg=styler, --style_fun=tidyverse_style]
+        args: [--style_pkg=styler, --style_fun=tidyverse_style,
+               --cache-root=styler-perm]
```
kaitejohnson, line 39: Is it worth updating what is in here on the cdcent template repo?

kaitejohnson (approved):
LGTM, just some clarifying questions

kaitejohnson:
@dylanhmorris I can merge this into #1, then I think I just need to fix the `pkgdown`

.pre-commit-config.yaml
dylanhmorris: Arguably but this isn't as critical as the use of `actions/cache`

dylanhmorris:
Sounds good will let you merge @kaitejohnson

PR #7, opened 2024-07-05
README and model definition draft

kaitejohnson:
Updated README.md and model_definition.md to describe the `wwinference` modeling package more generally. 

The README is still pretty limited, we can add to it as we add to the package.

I reorganized the model definition a bit to make the hierarchical structure more centered, since this is the only model we will provide to users (at least for now). We can add to this showing an option to fit the model without wastewater when we add this in #10


model_definition.md
```diff
@@ -0,0 +1,375 @@
...
+flowchart LR;
+    GlobalR --> GlobalI;
+    GlobalR --> LocalR;
+    LocalR --> GlobalR;
+    LocalR --> LocalIX;
+    LocalR --> LocalIY;
+    LocalIX --> LocalWWX;
+    LocalIY --> LocalWWY;
+    GlobalI --> Globalhosp;
+```
```
dylanhmorris, line 30: Don't you want to show `Global I` as the sum of the `Local I`, not as a function of `Global R`? What we have now looks like a model without bottom-up infection aggregation.

Also note that you can use $\TeX$-like math notation, so you can have e.g. $\mathcal{R}$ if you want.

dylanhmorris (commented):
One question about the model diagram.

model_definition.md
kaitejohnson: Ooh ok let me try to use the Latex formatting. I did think about that, I need to figure out how to put it in the diagram (maybe a box around the two Is with an arrow pointing to the Global I), and then remove the connection from global R to global I?

kaitejohnson: I tried using `Global$R$` and just got the $. I think @gvegayon said that github might not support this yet even though mermaid does?

gvegayon: That's right, last time I checked, GitHub doesn't support all features. I don't think it supports math yet, but it does markdown!:

```mermaid
flowchart LR;
  X["`This should be __bold__ and _cursive_`"]
  Y["This should print the delta symbol $$\delta$$"]
```

## Session info version

```mermaid
info
```

README.md
```diff
@@ -1,15 +1,80 @@
...
+See our Model Definition page for a mathematical description of the generative model, and the Getting Stated vignette to see an example of how to run the inference model on simulated data.
+
+The intention is for {wwinference} to provide a user-friendly R-package interface for running forecasting models that use wastewater concentrations combined with other more traditional epidemiological signals such as cases or hospital admissions. It aims to be a re-implementation of the modeling components contained in the [wastewater-informed-covid-forecasting](https://github.com/CDCgov/wastewater-informed-covid-forecasting) project repository, with
+an emphasis here on making it easier for users to supply their own data.
+
+We recommend reading the [model definition](model_definition.md) to learn more about how the model is structured and running the ["Getting Started" vignette](vignettes/wwinference.Rmd) for an example of how to fit the model to simulated data of COVID-19 hospital admissions and wastewater concentrations.
+This will help make clear the data requirements and how to structure this data to fit the model.
+
+## Project Admin
```
dylanhmorris, line 20: plural? 

```suggestion
## Project Admins
```

model_definition.md
dylanhmorris, line 30: Omit for now, add as a separate PR

```suggestion
```

dylanhmorris (changes requested):

dylanhmorris (approved):
LGTM!

PR #11, opened 2024-07-08
Port over the existing tests and set up CI

kaitejohnson:
This PR moves over the existing testing infrastructure from https://github.com/CDCgov/wastewater-informed-covid-forecasting, and modifies the tests as needed to fit the new functions. It also sets up CI to run the tests every time there is a PR into main or prod. Other scope creep things it does:
- adds documentation to the `data` using `data.R` 
- includes some but not complete set of examples for exported functions

Definitely need to figure out how to set up caching because the `r-cmd-check` is taking on the order of 20 min to fail :(  I think this can be a separate issue though. 

Do we want to set up any other CI? For example, I know @zsusswein set up code-cov (we would not do well right now....) and also a check to make sure that the change log( `NEWS.md`) was updated every time a PR merges to main 

Need to have main merged in once #5  is merged


kaitejohnson:
Fixing up a few things but think this is almost ready @dylanhmorris however, I don't the `pkgdown` is working correctly. When I go to `Settings >Pages` it says that it was deployed 2 weeks ago, and I think it should have been deployed with the latest merge to main?

kaitejohnson:
> Fixing up a few things but think this is almost ready @dylanhmorris however, I don't the `pkgdown` is working correctly. When I go to `Settings >Pages` it says that it was deployed 2 weeks ago, and I think it should have been deployed with the latest merge to main?

Edit: added `main` to `pkgdown.yaml`, and presuming this should fix it. This is ready for review. Note I moved the COVID delay distributions to `delay_distribs.R` to keep them as internal package functions.

R/postprocess.R
dylanhmorris, line 1: Looks like this file got accidentally reintroduced.

dylanhmorris (changes requested):

R/postprocess.R
kaitejohnson: Whoops, just fixed

kaitejohnson:
@dylanhmorris this is ready for re-review

R/get_stan_data.R
```diff
@@ -476,7 +476,7 @@ get_ww_values <- function(ww_data,
 
     # Get the vector of log wastewater concentrations
     log_conc <- ww_data |>
-      dplyr::mutate(log_conc = as.numeric(log(!!sym(
+      dplyr::mutate(log_conc = as.numeric(log(!!rlang::sym(
         ww_measurement_col_name
       ) + 1e-8))) |>
```
dylanhmorris, line 481: I'd use the `.data` pronoun here. NB [some R CMD check details](https://rlang.r-lib.org/reference/dot-data.html#where-does-data-live-)

```suggestion
      dplyr::mutate(log_conc = as.numeric(log(.data[[ww_measurement_col_name]] + 1e-8))) |>
```

Also, why is the `as.numeric()` call needed?

dylanhmorris: Would be good to make the small added on value (here `1e-8`) configurable, but that can be a separate PR.

R/wwinference.R
```diff
@@ -27,6 +27,10 @@
 #' user based on the date they are producing a forecast
 #' @param mcmc_options The MCMC parameters as defined using
 #' `get_mcmc_options()`.
+#' @param spec_inits Boolean indicating whether or not to specify the
+#' initialization of the sampler, default is `TRUE`, meaning that the
+#' initialization lists will be passed to the `cmdstanr::sample()`
+#' function
```
dylanhmorris, line 33: Passing `init = NULL` to the `sample()` method of a `CmdStanModel` object leads to random initis (see [docs](mdstanr/reference/model-method-sample.html).

So instead of having the if/else on `spec_inits` in your fitting code, I'd do something like:

```R
init_lists <- NULL
if(spec_inits) {
  init_lists <- c()
    init_lists[[i]] <- get_inits(stan_data, params)
  }
}
```

I'd also consider naming this flag something like `generate_initial_values`, since `spec_inits` slightly clashes with the `model_spec` argument.

inst/stan/simplewwinference.stan
dylanhmorris, line 1: Did you mean to reintroduce this?

dylanhmorris (changes requested):
A few things

DESCRIPTION
```diff
@@ -72,12 +72,19 @@ Imports:
     RcppTOML,
     cli,
     tibble,
+    usethis,
```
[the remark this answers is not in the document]
kaitejohnson: Argh my bad

DESCRIPTION
```diff
@@ -72,12 +72,19 @@ Imports:
...
     tidybayes,
     tidyr,
     purrr,
     withr,
     cmdstanr (>= 0.8.0),
     rlang,
-    scales
+    scales,
+    ggplot2,
+    rcmdcheck,
```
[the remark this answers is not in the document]
kaitejohnson: rcmdcheck? or scales?

DESCRIPTION
```diff
@@ -72,12 +72,19 @@ Imports:
...
+    posterior,
+    knitr
```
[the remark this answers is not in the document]
kaitejohnson: I think just the vignette

DESCRIPTION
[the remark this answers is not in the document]
dylanhmorris: `rcmdcheck`

data-raw/covid_pmfs.R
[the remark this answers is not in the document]
kaitejohnson: They're not exported package functions, just internal functions! My understanding was that everything in the `R/` can be used by its function name in the package, but only certain functions are "exported" to be on the user side?

R/get_stan_data.R
kaitejohnson: Hmm the `as.numeric()` was probably a holdover from a previous syntax that wasnt working properly. This makes sense.

DESCRIPTION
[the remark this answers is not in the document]
kaitejohnson: I think its the same thing as `data-raw/` I will add it to the exclusions in the `check R dependencies`

inst/stan/simplewwinference.stan
kaitejohnson: Whoops no!

data-raw/covid_pmfs.R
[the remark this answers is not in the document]
dylanhmorris: The gotcha here is that the data recipe `.R` scripts stored in `data-raw` are not themselves internal to the package, so they don't "know" about functions defined in `R/`

DESCRIPTION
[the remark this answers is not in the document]
dylanhmorris: okay, then it should be removed from `Depends:`

R/wwinference.R
```diff
@@ -27,6 +27,10 @@
 #' user based on the date they are producing a forecast
 #' @param mcmc_options The MCMC parameters as defined using
 #' `get_mcmc_options()`.
+#' @param generate_initial_values Boolean indicating whether or not to specify
+#' the initialization of the sampler, default is `TRUE`, meaning that the
+#' initialization lists will be passed to the `cmdstanr::sample()`
```
dylanhmorris, line 32:
```suggestion
#' the initialization of the sampler, default is `TRUE`, meaning that
#' initialization lists will be generated and passed as the `init` argument 
#' to the model object [`$sample()`][cmdstanr::model-method-sample] call.
```

dylanhmorris (commented):
Minor wording suggestion

dylanhmorris (approved):
Provided the CI checks pass this LGTM!

Issue #25, opened 2024-07-15
Forward simulation spatially correlated unadjusted reproduction number


cbernalz:
## Goal
We want to update the `generate_simulated_data.R` function in the R folder to implement a spatially correlated unadjusted reproduction number into it.  We will also add necessary functions to allow different correlation structures to be used.

## Context
The current model used in the `generate_simulated_data.R` function is listed in the model definitions in the repo.  The change we are making will be at the unadjusted reproduction level.  We will add the spatial correlations here with the following model : 

Let $k$ represent a site, from sites $1$ to $n_k$.  Let $R_k^u(t)$ be the unadjusted reproduction number at time $t$ at site $k$.  So lets look at the combined "state" $R^u(t)$ that contains an AR(1) correlation structure on the first differences that is currently being used.  So 

$$
\log R^u(t) \sim \text{Normal}(\log R^u(t - 1) + \beta(\log R^u(t - 1) - \log R^u(t - 2)), \sigma_r),
$$

or

$$
\log R^u(t) = \log R^u(t - 1) + \beta(\log R^u(t - 1) - \log R^u(t - 2)) + \sigma_r\text{Normal}(0,1).
$$

Now let $\delta_k(t)$ be the time-varying subpopulation effect on $\log R^u(t)$.  So we can look at the vector of all $\log R_k^u(t)$'s, so

$$
\begin{pmatrix}
        \log R_1^u(t)\\
	\vdots\\
	\log R_{n_{k}}^u(t)\\
\end{pmatrix} = \log R^u(t)
\begin{pmatrix}
	1\\
	\vdots\\
	1\\
\end{pmatrix} + \vec{\delta}(t).
$$

Where $\vec{\delta}(t) = \left(\delta_1(t),...,\delta_{n_k}(t)\right)^{T}$.  We also have that $\vec{\delta}(t) = \varphi_{R(t)}\vec{\delta}(t - 1) + \vec{\epsilon}$, where $\vec{\epsilon}$ $\sim$ $\text{Normal}_{n_k}(0,\mathbf{\Sigma\_{\epsilon}}).$  

This adds a spatial correlation and also the current AR(1) temporal correlation to this $\vec{\delta}(\cdot)$ random variable.  Like before, we have that $0 < \varphi_{R(t)} < 1$.

[...]

PR #26, opened 2024-07-16
Adding spatial components to spatial-main.

cbernalz:
This is regarding the issue here : https://github.com/CDCgov/ww-inference-model/issues/25#issue-2409527383 .


R/exponential_decay_corr_func.R
```diff
@@ -0,0 +1,50 @@
+#' Generates the correlation matrix for an exponential
+#' decay correlation function.
+#'
+#' @param corr_function_params Sturcuted as follows : {dist.matrix, phi, l}.
+#'
+#' @return Correlation matrix.
```
kaitejohnson, line 6: Perhaps also give a bit more of a desription of how this is formatted

R/exponential_decay_corr_func.R
```diff
@@ -0,0 +1,50 @@
+#' Generates the correlation matrix for an exponential
+#' decay correlation function.
+#'
+#' @param corr_function_params Sturcuted as follows : {dist.matrix, phi, l}.
```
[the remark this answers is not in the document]
kaitejohnson: And also mention that the defaults are `NULL` which will return a 1-diagonal matrix which corresponds to no correlation between sites

R/exponential_decay_corr_func.R
```diff
@@ -0,0 +1,50 @@
...
+      nrow(dist_matrix) == ncol(dist_matrix)
+  )
+
+  # presets
+  corr_matrix <- matrix(
+    data = 0,
+    nrow = nrow(dist_matrix),
+    ncol = ncol(dist_matrix)
+  )
+
```
kaitejohnson, line 41: I think you could write this simply as:

`corr_matrix  <- exp(-dist_matrix / phi)^l)`

because of how R handles matrices and scalars

R/exponential_decay_corr_func.R
```diff
@@ -0,0 +1,50 @@
...
+  for (i in seq_len(nrow(dist_matrix))) {
+    for (j in seq_len(ncol(dist_matrix))) {
+      d_ij <- dist_matrix[i, j]
+      corr_matrix[i, j] <- exp(-(d_ij / phi)^l)
+    }
+  }
+
+  return(corr_matrix)
```
kaitejohnson, line 49: Looking at this, it seems this would just return NULL if the default values are used. What does this mean in terms of the downstream functionality of the forward simulation? I am assuming we don't want to pass around a NULL matrix. 

We don't necessarily have to have default values for the function, especially if we want to create a switch for using the spatial correlation or not

R/generate_simulated_data.R
```diff
@@ -50,6 +50,15 @@
 #' LOD across sites
 #' @param input_params_path path to the toml file with the parameters to use
 #' to generate the simulated data
+#' #' @param corr_function Correlation function for spatial site correlations
```
kaitejohnson, line 53:
```suggestion
#' @param corr_function Correlation function for spatial site correlations, default is to use the `independence_corr_func`
```

R/generate_simulated_data.R
```diff
@@ -50,6 +50,15 @@
...
+#' @param corr_fun_params Parameters required for correlation function
```
kaitejohnson, line 54: Would add what the default is and its implications

kaitejohnson: Alternatively, I think we could and probably should just create an boolean switch that indicates whether or not to generate spatially correlated data, so we could add an argument that is `@param include_spatial_corr = TRUE` and then run those components if TRUE and not if FALSE?

R/generate_simulated_data.R
```diff
@@ -50,6 +50,15 @@
...
+#' @param phi_rt Coefficient for AR(1) temporal correlation on subpopulation
+#' deviations.
+#' @param sigma_eps Parameter for construction of covariance matrix of spatial
+#' epsilon.
+#' @param scaling_factor Scaling factor for aux site.
+#' @param aux_site_bool Boolean to use the aux site framework with
```
kaitejohnson, line 60: Why would we want to default this to FALSE?

R/generate_simulated_data.R
```diff
@@ -50,6 +50,15 @@
 #' LOD across sites
 #' @param input_params_path path to the toml file with the parameters to use
 #' to generate the simulated data
+#' #' @param corr_function Correlation function for spatial site correlations
+#' @param corr_fun_params Parameters required for correlation function
+#' @param phi_rt Coefficient for AR(1) temporal correlation on subpopulation
```
kaitejohnson, line 55: Should add the default values for these!

R/generate_simulated_data.R
```diff
@@ -291,6 +300,47 @@ generate_simulated_data <- function(r_in_weeks = # nolint
     )
   }
 
+  log_r_site <- spatial_rt_process(
+    log_r_state_week,
+    corr_function,
+    corr_fun_params,
```
kaitejohnson, line 306: Can you describe what this is doing by default?

R/generate_simulated_data.R
```diff
@@ -291,6 +300,47 @@ generate_simulated_data <- function(r_in_weeks = # nolint
...
+      data = 0,
+      nrow = 1,
+      ncol = ncol(log_r_site)
+    )
+    delta[1] <- rnorm(
+      n = 1,
+      mean = 0,
+      sd = sqrt(scaling_factor) * sigma_eps
+    )
+    n_t
```
kaitejohnson, line 327: Is this supposed to be here?

R/independence_corr_func.R
```diff
@@ -0,0 +1,24 @@
+#' Generates the correlation matrix for a non-spatial process.
+#'
+#' @param corr_function_params NULL
+#'
+#' @return Correlation matrix of diagonal ones.
+#'
+independence_corr_func <- function(
```
kaitejohnson, line 7: I see now, so essentially you made this function to handle the case when we don't want spatial correlation. I think we can specify this a little more simply by adding a switch, curious your thoughts @cbernalz @dylanhmorris

R/spatial_rt_process.R
```diff
@@ -0,0 +1,56 @@
...
+#' @param phi_rt Coefficient for AR(1) temporal correlation on
+#' subpopulation deviations.
+#' @param sigma_eps Parameter for construction of covariance matrix
+#' of spatial epsilon.
+#'
+#' @return A matrix for unadjusted Rt where rows are subpopulations
+#' and columns are time points.
+#'
+spatial_rt_process <- function(log_state_rt,
+                               corr_function,
```
kaitejohnson, line 16: This generality is really nice!

kaitejohnson (changes requested):
@cbernalz This looks great!! Just a few thoughts, curious what you think:
- I think in this `spatial-branch`, we want to make the default that we do incorporate the spatial correlation. So I think we'd want by default a distance matrix (can be any of the things you previously setup), that actually generates the exponentially decayed spatial R(t)s
- I'd suggest adding a switch in `generate_simulated_data.R()` that either does the original implementation with no spatial correlation or does what you have set up to do here. 

For a subsequent issue:
- the output of this function becomes package data in `data-raw/vignette_data.R`. I think for a separate PR/issue (after making the default have spatial correlation with a distance matrix) we should regenerate the package data using the spatially correlated data, and add a plot showing it in the pre-processing section (acknowledging that we haven't fit it with spatial correlation yet but thats ok!)

R/exponential_decay_corr_func.R
cbernalz: For this function, I added a stop function that doesn't let the function run if the values in corr_function_params are still NULL.  Is that fine?

R/generate_simulated_data.R
cbernalz: Using the default settings, this will just provide us with the uncorrelated Rt that was done before.

R/generate_simulated_data.R
cbernalz: I forgot to actually define that variable.

R/generate_simulated_data.R
kaitejohnson: I think we want to have the defaults here use the spatial component, and then handle the auxiliary site as you had written. And then we want to ensure that if we don't want the spatial component to be included, that the auxiliary site R(t) gets handled the way it currently does (treated as just another site)

R/generate_simulated_data.R
```diff
@@ -291,6 +340,47 @@ generate_simulated_data <- function(r_in_weeks = # nolint
...
+      data = 0,
+      nrow = 1,
+      ncol = ncol(log_r_site)
+    )
+    delta[1] <- rnorm(
+      n = 1,
+      mean = 0,
+      sd = sqrt(scaling_factor) * sigma_eps
+    )
+    n_t <- seq_len(ncol(log_r_site))
```
kaitejohnson, line 367: This is creating a vector when I think you want it to create a scalar equant to the number of weeks (in which case, I would just use the previously defined `n_weeks`

R/generate_simulated_data.R
```diff
@@ -364,7 +364,7 @@ generate_simulated_data <- function(r_in_weeks = # nolint
       mean = 0,
       sd = sqrt(scaling_factor) * sigma_eps
     )
-    n_t <- seq_len(ncol(log_r_site))
+    n_t <- ncol(log_r_site)
```
kaitejohnson, line 367: I think we should use `n_weeks` here rather than redefining a new variable! https://github.com/CDCgov/ww-inference-model/blob/d255645e96f5d48b09f15addbd63d2c3db0a87b7/R/generate_simulated_data.R#L169

R/exponential_decay_corr_func.R
kaitejohnson: Add a bit more of a description of what these parameters are!

R/generate_simulated_data.R
```diff
@@ -291,6 +340,47 @@ generate_simulated_data <- function(r_in_weeks = # nolint
...
 
+  log_r_site <- spatial_rt_process(
+    log_r_state_week,
+    corr_function,
+    corr_fun_params,
+    phi_rt,
+    sigma_eps
+  )
+  # Auxiliary Site
+  if (aux_site_bool) {
```
kaitejohnson, line 351: Just for my own clarity, if `use_spatial_corr = FALSE` whether or not the user specifies `aux_site_bool` as true or false doesn't matter, because it will get reassigned as FALSE, and the `independenc_corr_fun()` will spit out a diagnonal matrix of of `n_sites + 1` rows ?

kaitejohnson (changes requested):
This is super close-- just one more question and one suggestion re handling of the `n_t` variable

R/generate_simulated_data.R
cbernalz: Yes, but this can be changed.  I think it actually might be better to allow `aux_site_bool` to be true if `use_spatial_corr` is false.

R/generate_simulated_data.R
```diff
@@ -101,7 +101,7 @@
 #'   phi_rt = 0.6,
 #'   sigma_eps = sqrt(0.02),
 #'   scaling_factor = 0.01,
-#'   aux_site_bool = TRUE
+#'   aux_site_bool = FALSE
```
kaitejohnson, line 104: In the example, wouldn't we want the auxiliary site to be true because we are going to use the spatial correlation function for the ww sites?

kaitejohnson (commented):
Just a quick question about the example otherwise looks good!

kaitejohnson (dismissed):
LGTM, want to make a new issue that updates the package data for this branch? 
You can label it with the `spatial-ww` tag. Will let you merge and delete your feature branch after!

R/generate_simulated_data.R
kaitejohnson: I would commit this suggestion and then resolve the conversation :)

kaitejohnson:
Oh ok -- I just am going to suggest we add one more thing which is that we add one line to the `NEWS.md` describing what was done in this PR! So something like:
- implemented forward simulation of spatially correlated subpopulations given an adjacency matrix, to be added as an option in the `generate_simulated_data()` function

kaitejohnson (approved):
LGTM!

PR #29, opened 2024-07-17
Modularize the functions in the `generate_simulated_data()` function

### LINKED ISSUE #3: Break up `generate_simulated_data.R` function into more modular functions
## Goal
Currently one long linear function that steps through each component of the model to generate simulated hospital admissions and wastewater data. It is highly complex and hard to follow. It would be easier if each individual module called a function. That way, when you want to change how a module is handled, you can just write a new function for it. 

## Requirements
- [ ] break down getting global and local infection dynamics, generating expected hospital admissions, and generating expected wastewater concentrations into separate modular functions
- [ ] call these in the `generate_simulated_data.R` function

kaitejohnson:
This PR closes #3 and attempts to make the `generate_simulated_data()` function a series of calls to other functions that represent the critical model components, with the goal being that you can swap in and out different functions for different models (for example how to get the subpopulation R(t) estimates).


R/checkers.R
```diff
@@ -23,6 +23,91 @@ check_date <- function(df, max_date, call = rlang::caller_env()) {
...
+#' length
+#'
+#' @param r_in_weeks a vector indicating the R(t) in weeks
+#' @param ot integer indicating the observed time: length of hospital admissions
+#'  calibration time in days
+#' @param nt integer indicating the nowcast time: length of time between last
+#' hospital admissions date and forecast date in days
+#' @param forecast_horizon integer indicating the duration of the forecast in
+#' days e.g. 28 days
+#' @param call Calling environment to be passed to the type checker
```
dylanhmorris, line 36:
```suggestion
#' @param call Calling environment to be passed to `cli::cli_abort()` for traceback.
```

dylanhmorris: Is your goal with this pattern is to avoid the error message begin:
```
Error in `check_rt_length()`:
```
and instead have it begin:
```
Error in `<function that called check_rt_length()>()`:
```
?

kaitejohnson: I think so, so that in this case since I call this in the `generate_simulated_data()`  function it would tell you there was an error in that function if you passed in an R(t) of the wrong length that would read 
`"The weekly R(t) specifed isn't long enough to produce",
        "infections for the specified observed/calibration time (`ot`)",
        "nowcast period (`nt`), or forecast horizon (`forecast_horizon`)"`

dylanhmorris: That what I thought. In that case, I am fairly confident the pattern is right but that the documentation should be updated per my suggestion above, as what is written does not apply to this case.

R/checkers.R
```diff
@@ -23,6 +23,91 @@ check_date <- function(df, max_date, call = rlang::caller_env()) {
...
+  invisible()
+}
+
+#' Check that the specified site and lab indices line up
+#'
+#' @param site vector of integers indicating which site (WWTP) each separate
+#' lab-site observation comes frm
+#' @param lab ector of integers indicating which lab the lab-site observations
+#' come from
+#' @param call
```
dylanhmorris, line 91:
```suggestion
#' @param call Calling environment to be passed to [cli::cli_abort()] for traceback.
```

R/checkers.R
```diff
@@ -23,6 +23,91 @@ check_date <- function(df, max_date, call = rlang::caller_env()) {
...
+}
+
+#' Check that the sum of the wastewater site populations don't exceed
+#' the total population
+#'
+#' @param pop_size integer indicating the population size in the hypothetical
+#' state
+#' @param ww_pop_sites vector indicating the population size in the
+#' catchment area in each of those sites
+#' @param call Calling environment to be passed to the type checker
```
dylanhmorris, line 65:
```suggestion
#' @param call Calling environment to be passed to [cli::cli_abort()] for traceback. 
```

R/checkers.R
dylanhmorris, line 1: Can't straightforwardly make an inline suggestion for `check_elements_non_neg` as it's not part of the diffs but it also needs the documentation for the `call` argument fixed.

R/compile_model.R
```diff
@@ -7,6 +7,7 @@
 #' for additional source files to include. Within each stan file,
 #' use #include {path to your file with the `stan` directory}.stan
 #'
+#'
```
dylanhmorris, line 10:
```suggestion
```

dylanhmorris: I don't think this extra line needs to be added. Can revert.

dylanhmorris (commented):
Partial review to suggest fixes to docs

R/generate_simulated_data.R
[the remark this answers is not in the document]
dylanhmorris: I would also strongly suggest trying to reuse the helper/formatting functions defined in https://github.com/CDCgov/ww-inference-model/blob/main/R/get_stan_data.R wherever possible in this function. Among other things, that means the data simulation doubles as an extra check on the robustness of those helper functions.

R/generate_simulated_data.R
```diff
@@ -142,6 +139,14 @@ generate_simulated_data <- function(r_in_weeks = # nolint
...
 
+  # Get other variables needed for forward simulation ------------------------
+
+  # Get pop fractions of each subpop. There will n_sites + 1 subpops
+  pop_fraction <- c(
+    ww_pop_sites / pop_size,
+    (pop_size - sum(ww_pop_sites)) / pop_size
+  )
+  n_subpops <- length(pop_fraction)
   # Pull parameter values into memory
```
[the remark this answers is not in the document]
kaitejohnson: Fixed in #54 , will fix here as well

R/generate_simulated_data.R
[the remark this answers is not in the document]
kaitejohnson: Hmmm I'm not sure I am following @dylanhmorris. These functions are mostly binding a bunch of vectors together to form something that looks like input data tables, whereas those functions are doing the opposite -- getting individual components and indices from data tables

kaitejohnson:
@dylanhmorris This is ready for re-review. I could combine more things into larger wrapper functions, but my preference is that each individual component have its own function. It does still mean that the `generate_simulated_data()` function is large and verbose, but I don't think there's a great way to avoid this in a complex model.

R/generate_simulated_data.R
```diff
@@ -142,34 +139,36 @@ generate_simulated_data <- function(r_in_weeks = # nolint
...
+  pop_fraction <- c(
+    ww_pop_sites / pop_size,
+    (pop_size - sum(ww_pop_sites)) / pop_size
+  )
+  n_subpops <- length(pop_fraction)
 
   # Create a tibble that maps sites, labs, and population sizes of the sites
   n_sites <- length(unique(site))
-  site_lab_map <- data.frame(
+  site_lab_map <- create_site_lab_map(
```
gvegayon, line 154: I would argue it is a good practice to be verbose about the passed arguments. In other words, unless the function has a single argument, I would explicitly 
`arg1 = my_arg1`.

R/generate_simulated_data.R
```diff
@@ -215,268 +214,205 @@ generate_simulated_data <- function(r_in_weeks = # nolint
...
 
 
+  # Global adjusted R(t) --------------------------------------------------
+  # I(t)/convolve(I(t), g(t)) #nolint
+  # This is not used directly, but we want to have it for comparing to the
+  # fit.
+  rt <- calc_rt(new_i_over_n,
+    convolve_fxn = model$functions$convolve_dot_product,
+    generation_interval,
+    uot,
```
gvegayon, line 380: I suggest using named arguments.

R/generate_simulated_data.R
```diff
@@ -215,268 +214,205 @@ generate_simulated_data <- function(r_in_weeks = # nolint
...
-    dplyr::arrange(site, lab, date) |>
-    dplyr::select(date, site, lab, genome_copies_per_ml, lod, site_pop)
+  ww_data <- format_ww_data(
+    log_obs_conc_lab_site,
+    ot,
+    ht,
+    date_df,
+    site_lab_map,
+    lod_lab_site
+  )
```
gvegayon, line 395: Same here: named arguments.

R/generate_simulated_data.R
```diff
@@ -215,268 +214,205 @@ generate_simulated_data <- function(r_in_weeks = # nolint
...
-    ) |>
-    dplyr::select(
-      date,
-      daily_hosp_admits,
-      state_pop
-    )
+  hosp_data <- format_hosp_data(pred_obs_hosp,
+    dur_obs = ot,
+    pop_size,
+    date_df
```
gvegayon, line 402: And here.

R/generate_simulated_data.R
```diff
@@ -215,268 +214,205 @@ generate_simulated_data <- function(r_in_weeks = # nolint
...
 
-  # Make another one for model evaluation
-  hosp_data_eval <- tibble::tibble(
-    t = 1:(ot + ht),
-    daily_hosp_admits_for_eval = exp_obs_hosp,
-    state_pop = pop_size
+  hosp_data_for_eval <- format_hosp_data(pred_obs_hosp,
+    dur_obs = (ot + ht),
+    pop_size,
+    date_df
```
gvegayon, line 408: Same

R/model_component_fwd_sim.R
```diff
@@ -0,0 +1,472 @@
+#' Function to generate a global weekly R(t) estimate with added noise
+#'
+#' @param r_in_weeks The mean R(t) for each week of the simulation, can be
+#' longer than the 7*n_weeks
+#' @param n_weeks The number of weeks that we want to simulate the data for
+#' @param global_rt_sd The variation in the R(t) estimate to add
+#' intrinsic variability to the infection dynamics
+#'
+#' @return a weekly R(t) estimate with added noise
```
gvegayon, line 9: Since all these functions are associated, you can use the [`family` roxygen tag](https://roxygen2.r-lib.org/articles/index-crossref.html#family) to cross-reference them automatically. 

```suggestion
#' @family Model components
#' @return a weekly R(t) estimate with added noise
```

gvegayon (commented):
I think it looks good, @kaitejohnson! Fewer lines of code and more documentation! I have a handful of minor comments. Conditional on all tests passing, I would suggest you merge it.

R/generate_simulated_data.R
kaitejohnson: Ooh this is a good point, especially since there are a lot of these and order is constantly changing. Will go through and adjust so all of them are explicit

R/model_component_fwd_sim.R
kaitejohnson: Magic! So helpful thank you

NAMESPACE
gvegayon, line 1: Probably an unintended change, but you should not remove the `@export` tag. The `@family` tag doesn't export the function.

kaitejohnson: I actually did intend that-- I don't think I want these model components to be external user facing functions? Or maybe they should be?

kaitejohnson: I realized I had removed them in some components of the model but kept them in others

gvegayon: If it is internal, then there's no usage on the `family` tag. If you don't think others will find this useful, then sure, keep them internal :). When exposing functions, you have to think about the future release of the package: will it help the user?.

R/generate_simulated_data.R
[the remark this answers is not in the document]
dylanhmorris: Makes sense, thanks

NAMESPACE
kaitejohnson: I see! Ok think the family tip will be helpful for other things but probably not necessary here. Can go ahead and remove those tags!

gvegayon: And I would go further and add the `@noRd` tag.

kaitejohnson: That wouldn't create any of the man/files correct? I guess I think its nice to still have them if someone is in the weeds of the code and wants to know what the arguments are supposed to mean

kaitejohnson:
@dylanhmorris Let me know if this is good to merge, I made a separate issue for specifying all inputs https://github.com/CDCgov/ww-inference-model/issues/61

kaitejohnson:
@dylanhmorris This addresses #65 and updates the package data accordingly.

kaitejohnson:
@damonbayer This PR adds the `true_global_rt` dataframe to the package data. This contains the date and time columns and the `unadj_rt` and the `realized_rt`.

kaitejohnson:
@dylanhmorris @gvegayon Ready for re-review. 

I think it makes sense to make a separate issue to look into the back-calculation of R(t) and see why we are not able to reproduce a flat repeated value each day. I think it might have something to do with indexing/initialization but I'd want to look into it. Should be easier to do with a more functionalized version of the `generate_simulated_data()` function.

kaitejohnson:
Ok I think this fixes the back-calculation of R(t), by padding a 0 to the generation interval which replicates the forward

kaitejohnson:
@dylanhmorris Let me know if there are small changes needed to get this through review. Would be ideal to fix some of this before we have people using it, even if just prototyping. Apologies that it became such a large PR, but I think most of it has been reviewed by now :(

R/generate_simulated_data.R
```diff
@@ -215,272 +216,231 @@ generate_simulated_data <- function(r_in_weeks = # nolint
...
+    n_subpops = n_subpops,
+    r_weeks = unadj_r_weeks,
+    subpop_level_rt_variation = sigma_eps
+  )
 
-  new_i_over_n <- rep(0, (uot + ot + ht)) # State infections
-  for (i in 1:(n_sites + 1)) {
-    unadj_r_site <- ind_m %*% exp(log_r_site[i, ]) # daily R site
+  # Alternatively, can replace this with
+  # r_site <- spatial_rt_process(input_params) #nolint
```
dylanhmorris, line 275: I don't think this comment belongs in main
```suggestion
```

inst/stan/wwinference.stan
```diff
@@ -364,7 +364,10 @@ generated quantities {
   exp_state_ww_conc = exp(state_log_c);
 
   // Deterministic calculation of state level R(t) from incident infections
-  // and the generation interval
-  rt = (state_inf_per_capita ./ convolve_dot_product(state_inf_per_capita, gt_rev_pmf, uot + ot + ht))[uot+1: uot + ot + ht];
+  // and the generation interval, with a 0 added to account for the fact
+  // that our forward process assumes no contribution to onwards infections
+  // on the day of infection
+  rt = (state_inf_per_capita ./ convolve_dot_product(state_inf_per_capita,
+    reverse(append_row(0,generation_interval)), uot + ot + ht))[uot+1: uot + ot + ht];
```
dylanhmorris, line 371: I still would like this to be a bit more readable, but out of scope here.

data-raw/test_data.R
```diff
@@ -74,6 +74,7 @@ toy_stan_data <- wwinference::get_stan_data(
   inf_to_count_delay = model_spec$inf_to_count_delay,
   infection_feedback_pmf = model_spec$infection_feedback_pmf,
   params = model_spec$params,
+  include_ww = as.numeric(model_spec$include_ww),
```
dylanhmorris, line 77:
```suggestion
  include_ww = as.integer(model_spec$include_ww),
```
Since it's a flag in Stan, I'd coerfce to integer, not numeric.

inst/stan/wwinference.stan
kaitejohnson: I can make an issue to make a separate function for it like we have in `generate_simulated_data()` ?

R/model_component_fwd_sim.R
```diff
@@ -0,0 +1,484 @@
...
+get_global_rt <- function(r_in_weeks,
```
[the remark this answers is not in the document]
kaitejohnson: No the `r_In_weeks` is specified by the user, defaulted in natural scale. 

Good catch, added a manual replacement with a very small number.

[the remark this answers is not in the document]
dylanhmorris: Why not just add the noise on the transformed scale? Or not add noise at all?

R/model_component_fwd_sim.R
```diff
@@ -194,7 +199,9 @@ get_time_varying_daily_ihr <- function(p_hosp_mean,
 
 #' Get the predicted genomes per person in each subpopulation
 #'
-#' @param convolve_fxn function used to convolve infections with delay pmf
+#' @param convolve_fxn function used to convolve infections with delay pmf.
+#' This will typically take `convolve_dot_product()` from the
```
dylanhmorris, line 203:
```suggestion
#' This will typically take `convolve_dot_product()` function from the
```

R/model_component_fwd_sim.R
```diff
@@ -425,13 +432,15 @@ format_hosp_data <- function(pred_obs_hosp,
...
 #' infected on that day. This is a reasonable assumption, but to align the
-#' implementation of the forward process with our backword process, we have to
+#' implementation of the forward process with our backward process, we have to
 #' add a 0 density to day 0 of the passed in generation interval.
 #'
 #'
 #' @param new_i vector of numerics that spans the length of `tot_time`,
 #' representing the new incident infections per day
 #' @param convolve_fxn function used to convolve infections with delay pmf
+#' This will typically take `convolve_dot_product()` from the
```
dylanhmorris, line 442:
```suggestion
#' This will typically take the `convolve_dot_product()` function from the
```

R/model_component_fwd_sim.R
```diff
@@ -1,21 +1,24 @@
 #' Function to generate a global weekly R(t) estimate with added noise
 #'
-#' @param r_in_weeks The mean R(t) for each week of the simulation, can be
-#' longer than the 7*n_weeks
+#' @param r_in_weeks The mean R(t) for each week of the simulation in natural
+#' scale, can be longer than the 7*n_weeks
```
dylanhmorris, line 4:
```suggestion
#' scale, can be longer than the 7*n_weeks
```
Missed this before. Should the 7 be here?

dylanhmorris (commented):
A few final things.

inst/stan/wwinference.stan
dylanhmorris: sounds good.

R/model_component_fwd_sim.R
kaitejohnson: whoops no good catch

dylanhmorris (approved):
LGTM, thanks @kaitejohnson. Can you make a separate issue to reconsider the R(t) additional noise in the simulated data?

kaitejohnson:
> LGTM, thanks @kaitejohnson. Can you make a separate issue to reconsider the R(t) additional noise in the simulated data?

Sure, you mean to consider the noise being generated in log scale?

R/model_component_fwd_sim.R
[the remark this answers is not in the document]
kaitejohnson: Yeah, this is just basically to let the user specify something really boring and unrealistically smooth, and then add a bit of noise to it. I think its a bit easier to reason around this (particularly the right magnitude of the stdev) in natural scale. Can make a separate issue though.

dylanhmorris:
> > LGTM, thanks @kaitejohnson. Can you make a separate issue to reconsider the R(t) additional noise in the simulated data?
> 
> Sure, you mean to consider the noise being generated in log scale?

But also whether to have a user-specified timeseries with added noise at all (versus either/both of generating the $\mathcal{R}(t)$ from the underlying stochastic process or taking a user-specified timeseries as is (which the user can generate with whatever process they like)

kaitejohnson:
> > > LGTM, thanks @kaitejohnson. Can you make a separate issue to reconsider the R(t) additional noise in the simulated data?
> > 
> > 
> > Sure, you mean to consider the noise being generated in log scale?
> 
> But also whether to have a user-specified timeseries with added noise at all (versus either/both of generating the R ( t ) from the underlying stochastic process or taking a user-specified timeseries as is (which the user can generate with whatever process they like)

Yeah, I guess the "added noise" could certainly happen outside the function in any simulation analysis.

Issue #30, opened 2024-07-18
Updating package data in spatial-ww branch.


cbernalz:
## Goal
We want to use the spatial process that was recently added into the spatial-ww branch to update the package data that is used in the vignette.

## Requirements

- [ ] Update data using the `vignette_data.R` file

PR #31, opened 2024-07-18
30 updating package data in spatial ww branch

cbernalz:
Regarding issue https://github.com/CDCgov/ww-inference-model/issues/30#issue-2416516687 .


kaitejohnson (approved):
Confirming (for our own memory) that the `hosp_data.rda`, `hosp_data_eval.rda` and `ww_data.rda` were generated from running the `vignette_data.R` script after reinstalling the package and reloading the library. Same is true for the testing data, which is generated from running `test_data.R` and is saved to `sysdata.rda`

This LGTM! You can go ahead and merge

PR #33, opened 2024-07-18
Update pkgdown.yaml

kaitejohnson:
I think that we want to have `pkgdown` run whenever @cbernalz submits a PR into `spatial-main` and I think this sets that up.


dylanhmorris (approved):
LGTM

kaitejohnson:
@dylanhmorris I am still stuck on why the ruleset isn't picking up on the fact that `pkgodwn / buil (pull_request)` was successful?

kaitejohnson:
Nvm, seems to have work after rerunning

PR #34, opened 2024-07-22
2024-07-22 update : added stan spatial funcs.

cbernalz:


cbernalz:
I tested these functions within separate testfile that was not included in push.  All work as intended.

inst/stan/functions/spatial_rt_process_rng.stan
```diff
@@ -0,0 +1,37 @@
...
+
+  // presets
+  int n_time = dims(log_state_rt)[1];
+  matrix[n_sites,n_time] log_site_rt;
+  matrix[n_sites,n_time] delta;
+  vector[n_sites] eps_vec;
+
+  delta[,1] = delta_0;
+    for (t_i in 2:n_time) {
+      eps_vec = multi_normal_rng(rep_vector(0,n_sites), sigma_matrix_eps);
```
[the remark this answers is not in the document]
dylanhmorris: Suggest decomposing this function into two:

1. `construct_spatial_rt` that takes in an `eps_t_i` array (`n time x n_sites`)  and does the deterministic steps here
2. An `eps_vec_rng` that draws random `eps_t_i` arrays using `multi_normal_rng`

Can then create a `spatial_rt_process_rng` function that calls the two functions above under the hood.

inst/stan/functions/state_deviation_noise_vec_aux_rng.stan
```diff
@@ -0,0 +1,21 @@
...
+vector state_deviation_noise_vec_aux_rng(real scaling_factor,
+                                         real sigma_eps,
+                                         int n_time) {
+
+
+  vector[n_time] state_deviation_noise;
+
+    for (t_i in 1:n_time) {
+      state_deviation_noise[t_i] = normal_rng(0, scaling_factor * sigma_eps);
+
```
kaitejohnson, line 19:
```suggestion
}
```

kaitejohnson:
I know this is turning into a bear of a PR, but I think it would be helpful to see how these functions all come together if you called them in the `generate_simulated_data()` function. Also feel free to commit a `scratch/testing.R` folder file. I started a `scratch` folder in this next PR which I have been using to run the generation of simulated data line by line in a script format, as I find it easier to develop and troubleshoot that way. https://github.com/CDCgov/ww-inference-model/blob/3-modular-fxns-for-sim-data/scratch/sim_data_script.R

R/spatial_rt_process.R
```diff
@@ -29,13 +29,13 @@ spatial_rt_process <- function(log_state_rt,
   delta <- matrix(data = 0, nrow = n_subpopulations, ncol = n_time)
 
   # delta constr.
-  delta[, 1] <- mvrnorm(
+  delta[, 1] <- MASS::mvrnorm(
```
kaitejohnson, line 32: Just a note, always ok to do this IMO, but since you already have #' `@importFrom MASS mvrnorm` in`R/wwinference-package.R`, you don't HAVE to

inst/stan/functions/construct_aux_rt.stan
```diff
@@ -0,0 +1,30 @@
...
+  vector[n_time] log_aux_site_rt;
+  vector[n_time] state_deviation;
+
+  state_deviation[1] = state_deviation_init;
+  for (t_i in 2:n_time) {
+    state_deviation[t_i] = (state_deviation_ar_coeff*state_deviation[t_i-1])
+                              + state_deviation_noise_vec[t_i];
+  }
+
+  for (t_i in 1:n_time) {
```
kaitejohnson, line 25: Because these are all vectors and we aren't grabbing different indices, you can write this whole for loop as:
```suggestion
  log_aux_site_rt = log_state_rt + state_deviation
```

inst/stan/functions/construct_spatial_rt.stan
```diff
@@ -0,0 +1,37 @@
...
+
+  // presets
+  int n_time = cols(spatial_deviation_noise_matrix);
+  int n_sites = rows(spatial_deviation_noise_matrix);
+  matrix[n_sites,n_time] log_site_rt;
+  matrix[n_sites,n_time] spatial_deviation;
+
+  spatial_deviation[,1] = spatial_deviation_init;
+  for (t_i in 2:n_time) {
+    spatial_deviation[,t_i] = (spatial_deviation_ar_coeff*spatial_deviation[,t_i-1])
```
kaitejohnson, line 27: This looks great! Really nice job with passing in the deviation matrix and the initial error -- my only question is in the way we are naming things, which is a bit of semantics! 

$$
\begin{pmatrix}
        \log R_1^u(t)\\
	\vdots\\
	\log R_{n_{k}}^u(t)\\
\end{pmatrix} = \log R^u(t)
\begin{pmatrix}
	1\\
	\vdots\\
	1\\
\end{pmatrix} + \vec{\delta}(t).
$$

Where $\vec{\delta}(t) = \left(\delta_1(t),...,\delta_{n_k}(t)\right)^{T}$.  

We also have that $\vec{\delta}(t) = \varphi_{R(t)}\vec{\delta}(t - 1) + \vec{\epsilon}$, where $\vec{\epsilon}$ $\sim$ $\text{Normal}_{n_k}(0,\mathbf{\Sigma\_{\epsilon}})$  

In terms of this nomenclature, I see $ \varphi_{R(t)}$ which is your `spatial_deviation_ar_coeff` as still representing a temporal autocorrelation on the previous time step's deviation from the state mean -- that may or may not be informed by any spatial (e.g. distance or adjacency) matrix. For example, we could pass in a `spatial_deviation_noise_matrix` that is drawn from a correlation matrix that has no explicit distance/spatial dependence -- it just describes how correlated the sites are to one another. 

To give a very concrete example, say we had 3 sites in a state, 2 from big cities with a transit line between them and 1 from a rural town nearby one of the cities. In theory, we might expect that the spatial information is not the relevant component here, but that mobility patterns/flux in and out of the two cities make them more correlated in infection dynamics than the 1 rural area. 

I think I myself have been getting confused when we are actually referring to things that are spatially dependent e.g. functions of distance $d$ vs functions of the subpopulations ($k$)

I think this is a bit of semantics so feel free to leave as is in this PR and make a new issue! I think

inst/stan/functions/exponential_decay_corr_func.stan
```diff
@@ -0,0 +1,14 @@
+/**
+  * Assembles the correlation matrix with an exponential decay corr. func.
+  * @param dist_matrix matrix with distances.
+  * @param phi parameter for function.
+  * @param l paramter for function.
```
kaitejohnson, line 5:
```suggestion
  * @param l parameter for function.
```

inst/stan/functions/independence_corr_func.stan
```diff
@@ -0,0 +1,9 @@
+/**
+  * Assembles an independence correlation structure.
+  * @param n_sites number of sites/subpopulations.
+  * @return A correlation matrix for a spatially indpendent structure.
```
kaitejohnson, line 4:
```suggestion
  * @return A correlation matrix for a spatially independent structure.
```

inst/stan/functions/spatial_rt_process_rng.stan
```diff
@@ -0,0 +1,39 @@
...
+  //Presets
+  int n_time = dims(log_state_rt)[1];
+  int n_sites = rows(dist_matrix);
+
+  matrix[n_sites,n_time] spatial_deviation_noise_matrix = spatial_deviation_noise_matrix_rng(dist_matrix,
+                                                                                             sigma_eps,
+                                                                                             phi,
+                                                                                             l,
+                                                                                             n_time);
+  matrix[n_sites,n_time] log_site_rt = construct_spatial_rt(log_state_rt,
```
kaitejohnson, line 33: This looks really really nice! Great job breaking these into separate functions and separating out the sampling from the distributions from the deterministic transformations

inst/stan/functions/spatial_deviation_noise_matrix_rng.stan
```diff
@@ -0,0 +1,31 @@
+/**
+  * Generates a matrix where columns are random vectors for spatial process.
+  * Currently this is for an exponential correaltion function.
```
kaitejohnson, line 3:
```suggestion
  * Currently this is for an exponential correlation function.
```

inst/stan/functions/state_deviation_noise_vec_aux_rng.stan
```diff
@@ -0,0 +1,21 @@
+/**
+  * Generates a vector for state deviation noise for aux site.
+  * Currently this is for an exponential correaltion function.
```
kaitejohnson, line 3:
```suggestion
  * Currently this is for an exponential correlation function.
```

scratch/testfile.R
```diff
@@ -0,0 +1,484 @@
...
+      sd = initial_growth_prior_sd
+    )
+    # This is I0/N at the first unobserved time
+    log_i0_over_n_site[i] <- rnorm(
+      n = 1, mean = log_i0_over_n,
+      sd = 0.5
+    )
+  }
+
+  log_r_site <- spatial_rt_process_rng(
```
kaitejohnson, line 233: I am surprised you didn't need `model$functions$spatial_rt_process_rng` here!

kaitejohnson (approved):
@cbernalz This looks excellent. Great job with this! I'm really excited to put it into the model :) 

Two overall comments, that can be addressed both in separate PRs. 
1. I just want to consider us clarifying being very explicit about when we are talking about spatially dependent subppopulation correlations (e.g. a function of distance between sites, site adjacency) vs when are just talked about subpopulation correlations (that could be inferred or set from for example other types of data other than spatial information). Perhaps I am being pedantic though, but I think this has been tripping me up a bit in understanding the decomposition and when it is or is not spatially dependent.
2. The `testfile.R` is great! I would say for a second PR, you should incorporate the generation of the spatially correlated data using the stan functions directly into the `generate_simulated_data.R`. And in doing so, instead of using the `_rng` functions go ahead and separate the process of sampling from an MVN with doing the transformations given the "noise matrix". This will exactly mirror what you will want to do in the `wwinference.stan` model (they could even be a part of the same PR if you wanted to, though I would personally keep them separate). 

@dylanhmorris Let me know if you agree on next steps/ have any thoughts on my first point

inst/stan/functions/construct_spatial_rt.stan
athowes: e.g. $\mathbf{x} \sim \mathcal{N}(\mathbf{0}, \mathbf{I})$ where $\mathbf{x} = (x_1, \ldots, x_n)$ and $1, \ldots, n$ are indices for areas would be spatial random effects but unstructured.

scratch/testfile.R
cbernalz: I used the expose functions, but next time I will use that because there was a couple of naming conflicts with stan and r functions having the same names so I renamed the r functions that were causing the problem with `_r.R`.

inst/stan/functions/construct_spatial_rt.stan
dylanhmorris: Edited @kaitejohnson comment to fix a typo that was messing up $\TeX$ rendering

inst/stan/functions/construct_spatial_rt.stan
```diff
@@ -0,0 +1,37 @@
...
+                              + spatial_deviation_noise_matrix[,t_i];
+  }
+
+  for (t_i in 1:n_time) {
+    log_site_rt[,t_i] = (log_state_rt[t_i]*rep_vector(1,n_sites))
+                        + spatial_deviation[,t_i];
+  }
```
dylanhmorris, line 34: If we're not going to return the `spatial_deviation` matrix, perhaps just have `spectial_deviation_t_i` as a throwaway vector that gets repeatedly overwritten within the for-loop?

i.e. something like

```stan
  vector[n_sites] spatial_deviation_t_i = spatial_deviation_init;

  for (t_i in 1:n_time) {
    spatial_deviation_t_i = (spatial_deviation_ar_coeff*spatial_deviation_t_i)
    log_site_rt[,t_i] = (log_state_rt[t_i]*rep_vector(1,n_sites))
                        + spatial_deviation_t_i;
}
```
etc

inst/stan/functions/construct_aux_rt.stan
```diff
@@ -0,0 +1,30 @@
...
+    log_aux_site_rt[t_i] = log_state_rt[t_i] + state_deviation[t_i];
+  }
+
+  return log_aux_site_rt;
+}
```
dylanhmorris, line 28: As below with the `construct_spatial_rt` function, I might treat `state_deviation_t_i` as a throwaway variable.

inst/stan/functions/exponential_decay_corr_func.stan
dylanhmorris: Looks like this didn't get fixed? Also, would suggest being a bit more verbose here, e.g.:

```suggestion
  * @param l power to which to which to raise dist_matrix / phi before negating and exponentiating. l = 1 yields standard exponential decay.```

File: inst/stan/functions/exponential_decay_corr_func.stan [hunk 14]
```diff
@@ -0,0 +1,14 @@
+/**
+  * Assembles the correlation matrix with an exponential decay corr. func.
+  * @param dist_matrix matrix with distances.
+  * @param phi parameter for function.
```
dylanhmorris, line 4: Suggest being more verbose
```suggestion
  * @param phi scalar parameter controlling the decay rate of correlation with distance. Should be non-negative. Larger values imply a greater reduction in correlation with increased distance. 
```

inst/stan/functions/independence_corr_func.stan
```diff
@@ -0,0 +1,9 @@
...
+  */
+matrix independence_corr_func(int n_sites) {
+
+  return diag_matrix(rep_vector(1.0, n_sites));
```
dylanhmorris, line 8: Stan has an `identity_matrix` function:

```suggestion
  return identity_matrix(n_sites);
```

inst/stan/functions/spatial_deviation_noise_matrix_rng.stan
```diff
@@ -0,0 +1,31 @@
...
+  * @param dist_matrix Distance matrix for correlation constr.
+  * @param sigma_eps Parameter for construction of covariance matrix
+  * of spatial epsilon.
+  * @param phi Phi parameter for exponential decay corr. func.
+  * @param l l paraamter for exponential decay corr. func.
+  * @param n_time timepoints for number of random vectors of spatial process desired.
+  * @return A matrix for spatial deviation noise in use of spatial Rt process.
+  */
+matrix spatial_deviation_noise_matrix_rng(matrix dist_matrix,
```
dylanhmorris, line 12: If these function is going to be specifically for exponential decay, should indicate that in the function name. Alternatively (and I think this is my preference), you could rewrite this as a more generic function that takes in an abitrary`sigma_matrix`. One could then compute the `sigma_matrix` first and then pass it to this function.

If you wanted an exponential decay rng, you could then construct it by calling this more generic function with a `sigma_matrix` computed using exponential decay.

inst/stan/functions/spatial_deviation_noise_matrix_rng.stan
```diff
@@ -0,0 +1,31 @@
+/**
+  * Generates a matrix where columns are random vectors for spatial process.
+  * Currently this is for an exponential correaltion function.
+  * @param dist_matrix Distance matrix for correlation constr.
+  * @param sigma_eps Parameter for construction of covariance matrix
+  * of spatial epsilon.
+  * @param phi Phi parameter for exponential decay corr. func.
+  * @param l l paraamter for exponential decay corr. func.
```
dylanhmorris, line 8: Suggest changing these as above.

inst/stan/functions/spatial_rt_process_rng.stan
```diff
@@ -0,0 +1,39 @@
...
+  * of spatial epsilon.
+  * @param phi Parameter for construction of correaltion matrix.
+  * @param l Parameter for construction of correlation matrix.
+  * @param spatial_deviation_ar_coeff Coefficient for AR(1) temporal correlation on
+  * subpopulation deviations.
+  * @param spatial_deviation_init Initial vector of spatial deviation.
+  * @return A matrix for unadjusted Rt where rows are subpopulations
+  * and columns are time points.
+  */
+matrix spatial_rt_process_rng(vector log_state_rt,
```
dylanhmorris, line 15: As above, should indicate in the function name that this is for an exponential decay correlation or rewrite to be more generic and accept `sigma_matrix` as a pre-computed argument.

inst/stan/functions/state_deviation_noise_vec_aux_rng.stan
dylanhmorris: Again, looks like this wasn't fixed.

inst/stan/wwinference.stan
```diff
@@ -6,6 +6,14 @@ functions {
...
 #include functions/observation_model.stan
 #include functions/utils.stan
+#include functions/exponential_decay_corr_func.stan
+#include functions/independence_corr_func.stan
+#include functions/construct_spatial_rt.stan
+#include functions/spatial_deviation_noise_matrix_rng.stan
+#include functions/spatial_rt_process_rng.stan
+#include functions/construct_aux_rt.stan
+#include functions/state_deviation_noise_vec_aux_rng.stan
+#include functions/aux_site_process_rng.stan
```
dylanhmorris, line 16: If these aren't used in the actual model yet, I think best not to `#include` them here. I'm guessing the goal is to expose them to R, I would suggest for now writing a general `functions.stan` file that includes them all and using `cmdstanr`'s `$expose_functions()` functionality on that file.

dylanhmorris (changes requested):
This is shaping up very nicely, @cbernalz, thanks! A few comments/suggestions/questions.

R/generate_simulated_data.R
```diff
@@ -100,7 +104,9 @@
 #'   ),
 #'   phi_rt = 0.6,
 #'   sigma_eps = sqrt(0.02),
+#'   spatial_deviation_init = c(0.09, 0.32, 0.09, -0.78, -0.24, 0.16),
```
kaitejohnson, line 107: I might suggest, since these should be random variables, that you actually draw this and the `state_deviation_init` within this function rather than passing them in

R/generate_simulated_data.R
```diff
@@ -339,41 +348,29 @@ generate_simulated_data <- function(r_in_weeks = # nolint
...
 
-  log_r_site <- spatial_rt_process(
-    log_r_state_week,
-    corr_function,
-    corr_fun_params,
-    phi_rt,
-    sigma_eps
+  # Using stan exposed functions for forward spatial Rt process.
+  sigma_matrix <- sigma_eps * corr_function(corr_fun_params)
+  spatial_deviation_noise_matrix <- spatial_deviation_noise_matrix_rng(
```
kaitejohnson, line 353: I think you want to do a similar thing with the `spatial_deviation_init` (though it doesn't need to call a stan function, just use R to draw from a normal with mean `log_r_state_week` and stdeviation `sigma_rt`

inst/stan/functions/spatial_deviation_noise_matrix_rng.stan
```diff
@@ -1,30 +1,17 @@
 /**
   * Generates a matrix where columns are random vectors for spatial process.
-  * Currently this is for an exponential correaltion function.
-  * @param dist_matrix Distance matrix for correlation constr.
-  * @param sigma_eps Parameter for construction of covariance matrix
-  * of spatial epsilon.
-  * @param phi Phi parameter for exponential decay corr. func.
-  * @param l l paraamter for exponential decay corr. func.
+  * @param Sigma_matrix Matrix of covariances, constructed externally.
```
kaitejohnson, line 3: Would still name this `sigma_matrix` bc of linting...

kaitejohnson: The linter will expect that everything will be `snake_case`. I think it only checks the R files but when you pull this in as a variable in R will want it to not be flagged

inst/stan/functions/spatial_functions.stan
```diff
@@ -0,0 +1,12 @@
+/**
+  * File to call all desired spatial functions for use with expose_functions()
```
kaitejohnson, line 2: Did you use this in `generate_simulated_data.R` instead of calling `wwinference.stan`? If so, I must have missed it

kaitejohnson (commented):
A few minor comments -- getting very close! I think we should stick with the same convention as the rest of `generate_simulated_data()` and we should draw from the distribution within the function to generate the deviation/error vectors/ values/matrices

inst/stan/functions/spatial_functions.stan
cbernalz: I used it to reduce the functions used in the function block of wwinference.stan.  This is regarding Dylans last request from above.

kaitejohnson: But so you'll want to replace:
```
  model <- cmdstanr::cmdstan_model(
    stan_file = system.file(
      "stan", "wwinference.stan",
      package = "wwinference"
    ),
    compile = TRUE,
    compile_standalone = TRUE,
    force_recompile = TRUE
  )
  model$expose_functions(global = TRUE)
```

With 
  model <- cmdstanr::cmdstan_model(
    stan_file = file.path("inst", "stan","functions", "spatial_functions.stan"),
    compile = TRUE,
    compile_standalone = TRUE,
    force_recompile = TRUE
  )
  model$expose_functions(global = TRUE)

R/generate_simulated_data.R
cbernalz: Are you referring to the spatial_deviation_init or the whole spatial_deviation_noise_matrix?  I am a bit confused on this comment.

kaitejohnson: I am suggesting that instead of passing in `spatial_deviation_init`, you do the same thing that you do here for `spatial_deviation_noise_matrix`. So to get the initial deviaiton you would do something like 

`spatial_deviation_init <- log_state_Rt_week[,1] + rnorm(n= n_weeks, mean = 0, sd = sigma_rt)`

cbernalz: After my one-on-one meeting with @dylanhmorris, we decided to not have an init value for either spatial or aux site rt functions.  For the initial values, we will just use $\text{Normal}\_{n\_{k}}(0,\Sigma)$ and $\text{Normal}(0,\text{ScalingFactor}\times\sigma\_{\epsilon})$ for spatial and aux rt's.  This will be used in the construct functions after passing in a matrix of random vectors for spatial rt and after passing in a vector of random values for aux site.  If we end up deciding to do something different with the initial values of the matrix of random vectors or vector of random values, we can add that change in the construction of these random quantities, within the noise rng functions.

kaitejohnson: I think that makes sense, these changes look good!

kaitejohnson (dismissed):
These changes look good. Pending CI passes and also I would run and generate the simulated data using these functions just as a check, this looks good to me.

kaitejohnson (approved):
LGTM, good catch on not committing the compiled stan model!

kaitejohnson (changes requested):
See if that works!

R/generate_simulated_data.R
```diff
@@ -198,6 +199,16 @@ generate_simulated_data <- function(r_in_weeks = # nolint
     force_recompile = TRUE
   )
   model$expose_functions(global = TRUE)
+  spatial_fxns <- cmdstanr::cmdstan_model(
+    stan_file = system.file(
+      "stan", "spatial_functions.stan",
```
kaitejohnson, line 204:
```suggestion
      "stan", "functions", "spatial_functions.stan",
```

kaitejohnson (approved):

kaitejohnson:
@dylanhmorris Are there other changes you requested that haven't been addressed?

dylanhmorris:
Reviewing now.

inst/stan/functions/state_deviation_noise_vec_aux_rng.stan
```diff
@@ -0,0 +1,20 @@
...
+                                         real sigma_eps,
+                                         int n_time) {
+
+  vector[n_time] state_deviation_noise;
+
+  for (t_i in 1:n_time) {
+      state_deviation_noise[t_i] = normal_rng(0, scaling_factor * sigma_eps);
+  }
+
+  return state_deviation_noise;
```
dylanhmorris, line 19: `normal_rng` is vectorized, so I believe this works:

```suggestion
  vector[n_time] state_deviation_noise = normal_rng(rep_vector(0, n_time), scaling_factor * sigma_eps);

  return state_deviation_noise;
```

dylanhmorris (commented):
Minor suggestions. Thanks, @cbernalz!

dylanhmorris (approved):
LGTM, thanks @cbernalz!

dylanhmorris (approved):

PR #37, opened 2024-07-29
Input hospital admissions and wastewater data checks 

### LINKED ISSUE #4: Add validation and checks for all input data
## Goal
Want to make sure that the input data that a user provides for hospital admissions and wastewater datasets meets requirements needed for the model functions to work correctly. If not, we want to provide them with coherent error messaging to tell them how to modify it. 

## Requirements
- [ ] model some checks after @zsusswein 's `checkers.R` https://github.com/CDCgov/cfa-gam-rt/blob/main/R/checkers.R and other validations to ensure that input data is formatted properly.

kaitejohnson:
Here are some validation checks that user-provided input data (for wastewater and hospital admissions) are formatted correctly. The overall goal is to provide really user-friendly error messaging to let the user know when something they have passed in will be incompatible/incoherent with downstream functionality. In the case of this project, we need a very specific structure for the current implementation of the model (e.g. one hospital admissions dataset from a global population, one or more wastewater concentration datasets with lab and site identifiers), so there are pretty strict requirements for how the input data needs to be formatted.  The goal here is to fail early and loudly, and tell the user what aspect of their data is causing the failure.

What this does so far:
For the wastewater data:
- checks we have the required column names
- checks that the columns are dates, the values are non-negative, the site and lab identifiers are integers or characters, the site population sizes are integers, checks that there are no missing values (for the ww data, we don't want NA padding)
For the hospital admissions data:
- checks we have the required column names
- checks that the columns are dates, that there is only a single population size for the data, checks that there is only one observation per day, checks that counts are positive and integers 

We will want to go through and add more of these, in particular checks that look at both input data sets and say give a warning if the sum of the populations of the sites are greater than the global population. 

I relied heavily on @zsusswein's [gam package](https://github.com/CDCgov/cfa-gam-rt/blob/main/R/checkers.R), and then created a bunch of more very specific to this model/project checks as well. 

@seabbs @dylanhmorris Tagging you for any thoughts you might have. This is complete but I think closes #4


cbernalz (approved):

R/checkers.R
```diff
@@ -45,3 +56,345 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+  if ((!rlang::is_integerish(x)) || (!inherits(x, "Date"))) {
+    throw_type_error(
+      object = x,
+      arg_name = arg,
+      expected_type = "Date",
+      call = call
+    )
+  }
+  invisible()
+}
```
dylanhmorris, line 77: I'm surprised we need to create all these type checkers for ourselves. Is there not a standard tidyverse or tidyverse-adjacent library for this? 

cc @seabbs

NAMESPACE
```diff
@@ -68,6 +68,7 @@ importFrom(ggplot2,theme)
 importFrom(lubridate,ymd)
 importFrom(posterior,as_draws_list)
 importFrom(posterior,subset_draws)
+importFrom(rlang,abort)
```
dylanhmorris, line 71: where is this used? I only see `cli::cli_abort()` calls (note that `cli::cli_abort()` calls `rlang::abort()` under the hood, but you shouldn't need to import the name.

dylanhmorris (commented):
Some questions @kaitejohnson cc @seabbs

NAMESPACE
kaitejohnson: I can try removing it? I don't think it is used anywhere

R/checkers.R
zsusswein: You may want [checkmate ](https://mllg.github.io/checkmate/)

NAMESPACE
kaitejohnson: Oh wait this gets generated automatically, so it must be in here... let me look

kaitejohnson: Ok nope this is what I get for copying things from other people without fully understanding them :) @zsusswein has set up his package functions to explicitly do the importfrom in the function, and had `@importFrom rlang abort` in his function that I modified

R/checkers.R
```diff
@@ -45,3 +56,344 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+#' additional model functionality has been added.
+#'
+#'
+#' @param x the vector to check
+#' @param arg the name of the vector to check
+#' @param call Calling environment to be passed to [cli::cli_abort()] for
+#' traceback.
+#'
+#' @return NULL, invisibly
+check_for_repeat_dates <- function(x, arg = "x", call = rlang::caller_env()) {
```
dylanhmorris, line 380: this doesn't seem specific to dates in the implementation. Seems like it is best done with something like [`assertr`](https://docs.ropensci.org/assertr/articles/assertr.html#assertrs-predicates)'s uniqueness checking

dylanhmorris (changes requested):
I think we should use leverage existing tools here.

R/checkers.R
kaitejohnson: Definitely! I didn't realize this existed, it looks very useful!

kaitejohnson: Ok my only complaint/thought here is by using these, I don't see an argument to pass a more helpful errror message (e.g. based on the `arg`). So as written, it will tell you which argument failed the test, but I think if I remove the `check_{date/vectors/nonmissinginess}` I lose this functionality and will just get an error about that type but not specific to the `arg`

kaitejohnson:
Ok @dylanhmorris I made the suggested edits in the `validate.R` files, using `checkmate::check_x()` and `assertr::fxn()` where present. My one question that I have is would it be terribly redundant to instead still wrap these in the hand written `check_element()` functions, so that for example for the type checkers it will tell you what argument failed the type checker rather than just that it failed, and for the ones that are very specific to our model, we can give more verbose error messages? 

For example, while the `assertr::not_na()` will error if I pass in a vector containing NAs, as written originally, I will get an error and I will get this error message:
```
c("{.arg {arg}} has missing values",
        "i" = "Missing values are not supported in {.arg {arg}}",
        "!" = "Missing element(s) index: {.val {which(is_missing)}}"
      ),
```
Whick maybe is unnecessary but is a bit more of an explanation. So for the wastewater concentration column, we would be able to have something that tells them they can't have missing values in their wastewater concentration data and which element its missing in, which seems useful? 

Maybe am overthinking here!

R/checkers.R
```diff
@@ -186,58 +186,42 @@ throw_type_error <- function(object,
...
+    must.include = expected_col_names
+  ))
+  if (class(res) == "try-error") {
     cli::cli_abort(
-      c("`date` column missing from input wastewater data"),
+      c("Required columns are missing from the input wastewater data"),
       class = "wwinference_input_data_error",
       call = call
     )
   }
```
dylanhmorris, line 222: If you want to set a custom error message, the simplest thing to do here is use [`checkmate::check_names()`](https://mllg.github.io/checkmate/reference/checkNames.html) (which returns `TRUE` on check success and `FALSE` on check failure) rather than using `checkmate::assert_names` and a try/catch.

```suggestion
  if(!checkmate::check_names(column_names, must.include = expected_col_names)) {
    cli::cli_abort(
      c("`date` column missing from input wastewater data"),
      c("Required columns are missing from the input wastewater data"),
      class = "wwinference_input_data_error",
      call = call)
  }

File: R/checkers.R (hunk 1 above)
Comment (dylanhmorris) [reply R/checkers.R:77]: The `assert_` functions from `checkmate` give you sume but not total opportunity to customize error messages. if you want more control over the message you can use `check_*()` rather than `assert_*()` and then raise the errors yourself.

**Comment by dylanhmorris** [comment]
For `assertr` I would suggest trying to configure things via this functionality: https://docs.ropensci.org/assertr/articles/assertr.html#success-error-and-defect-functions

File: R/checkers.R [hunk 5]
```diff
@@ -45,3 +56,213 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+      conc_col_name
+    },
+    {
+      lod_col_name
+    },
+    add_req_col_names
+  )
+
+  # This tells you whats missing
+  checkmate::check_names(column_names, must.include = expected_col_names)
```
kaitejohnson, line 156: @dylanhmorris I think this would be where I would want to maybe submit an issue to `checkmate` because we don't want to have to assert this twice -- the first gives you a useful error message about what names are missing, but I also want a custom error message that puts it in context, e.g. what data are we applying this to (which I have added below).

R/checkers.R
```diff
@@ -45,3 +56,213 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+  if (!checkmate::check_names(
+    column_names,
+    must.include = expected_col_names
+  )) {
+    cli::cli_abort(
+      "Required columns are missing from the input wastewater data",
+      class = "wwinference_input_data_error",
+      call = call
+    )
+  }
```
dylanhmorris, line 168:
```suggestion
  name_check_result <- checkmate::check_names(column_names, must.include = expected_col_names)
  if (!name_check_result) {
    cli::cli_abort(
      "Required columns are missing from the input wastewater data",
      name_check_result,
      class = "wwinference_input_data_error",
      call = call
    )
  }
```

R/checkers.R
dylanhmorris: https://github.com/CDCgov/ww-inference-model/pull/37/files#r1702288282

R/checkers.R
```diff
@@ -1,12 +1,22 @@
 #' Check that all dates in dataframe passed in are before a specified date
 #'
+#' @description
+#' This function is specifically meant to ensure that the data in the `df`
+#' specified does nto contain data time-stamped beyond the `max_date`. The
```
dylanhmorris, line 5:
```suggestion
#' specified does not contain data time-stamped beyond the `max_date`. The
```

R/checkers.R
```diff
@@ -1,12 +1,22 @@
...
+#'
+#'
 #' @param df dataframe with `date` column
 #' @param max_date string indicating the maximum date in ISO8601 convention
 #' e.g. YYYY-MM-DD
 #' @param call Calling environment to be passed to the type checker
 #'
 #' @return NULL, invisibly
-check_date <- function(df, max_date, call = rlang::caller_env()) {
+check_date_logic <- function(df, max_date, call = rlang::caller_env()) {
```
dylanhmorris, line 19: Should this definitely take in the whole data frame versus take in that columns, i.e.

```suggestion
check_date_logic <- function(date_vector, max_date, call = rlang::caller_env()) {
```

R/checkers.R
dylanhmorris: I do agree that it would be nice to be able to customize the error message directly in `assert_`.

R/checkers.R
```diff
@@ -45,3 +56,213 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+    must.include = expected_col_names
+  )) {
+    cli::cli_abort(
+      "Required columns are missing from the input count data",
+      class = "wwinference_input_data_error",
+      call = call
+    )
+  }
+
+  invisible()
```
dylanhmorris, line 221: as above https://github.com/CDCgov/ww-inference-model/pull/37/files#r1702288282

R/checkers.R
kaitejohnson: This makes sense!

R/compile_model.R
```diff
@@ -7,6 +7,7 @@
 #' for additional source files to include. Within each stan file,
 #' use #include {path to your file with the `stan` directory}.stan
 #'
+#'
```
dylanhmorris, line 10:
```suggestion
```

R/preprocessing.R
```diff
@@ -18,20 +18,34 @@
...
 preprocess_ww_data <- function(ww_data,
                                conc_col_name = "genome_copies_per_ml",
                                lod_col_name = "lod") {
+  # This checks that we have all the right column names
+  check_required_ww_inputs(
+    ww_data,
+    conc_col_name,
+    lod_col_name
+  )
+  # This checks that the contents of the wastewater concentration data is valid
```
dylanhmorris, line 41: Can we make this comment more informative or delete it?

R/preprocessing.R
```diff
@@ -18,20 +18,34 @@
...
-#'                                            lod_col_name = "lod"
-#'                                            )
+#'   conc_col_name = "conc",
+#'   lod_col_name = "lod"
+#' )
 preprocess_ww_data <- function(ww_data,
                                conc_col_name = "genome_copies_per_ml",
                                lod_col_name = "lod") {
+  # This checks that we have all the right column names
+  check_required_ww_inputs(
```
dylanhmorris, line 36: This comment suggests the function could be more meaningfully named. Maybe `check_all_needed_ww_columns_present`?

R/checkers.R
```diff
@@ -45,3 +56,209 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+  if (!(rlang::is_integerish(x) || rlang::is_character(x))) {
+    throw_type_error(
+      object = x,
+      arg_name = arg,
+      expected_type = "integer or character",
+      call = call
+    )
+  }
+  invisible()
+}
```
dylanhmorris, line 85: This should be done with checkmate functions as well. As below, can either combine assertions like this:

`checkmate::assert(checkmate::check_integerish(x), checkmate::check_character(x))`

or can use `checkmate::check_*()` output in `if` / custom `cli::cli_abort()` calls.

dylanhmorris (commented):
Some more partial comments

R/checkers.R
kaitejohnson: I guess since we have the other checks for the date column being present this is fine!

R/preprocessing.R
kaitejohnson: Sure can change to this for both colname checks

kaitejohnson:
@dylanhmorris Let me know what you think of the changes I just made. 

I removed the dependency on `assertr` because as is I don't think we are using it to DRY-ify any code (e.g. we aren't using it to string together assertions or more generally apply to an entire dataframe). Instead, we have "check" functions using `checkmate` and baser/dplyr that do the specific checking on a specified vector, and then we bind wrap these together in the validate functions. I like the current functionality because passing the arg returns a helpful error message that indicates what column in the dataframe is erroring and even what index it is at.

R/checkers.R
```diff
@@ -45,3 +57,268 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+#'
+#' @param x the vector to check
+#' @param arg the name of the vector to check
+#' @param call Calling environment to be passed to [cli::cli_abort()] for
+#' traceback.
+#'
+#' @return NULL, invisibly
+check_non_missingness <- function(x, arg = "x", call = rlang::caller_env()) {
+  any_missing <- any(is.na(x))
+  if (any_missing) {
```
dylanhmorris, line 71:
```suggestion
  if (checkmate::anyMissing(x)) {
```

R/checkers.R
```diff
@@ -45,3 +57,268 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+
+#' Check there is no missignness in a particular vector
+#'
+#' @param x the vector to check
+#' @param arg the name of the vector to check
+#' @param call Calling environment to be passed to [cli::cli_abort()] for
+#' traceback.
+#'
+#' @return NULL, invisibly
+check_non_missingness <- function(x, arg = "x", call = rlang::caller_env()) {
```
dylanhmorris, line 69: For consistency with `checkmate` these functions should be named `assert_` rather than `check_`
```suggestion
assert_non_missingness <- function(x, arg = "x", call = rlang::caller_env()) {
```

R/checkers.R
```diff
@@ -45,3 +57,268 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
   }
   invisible()
 }
+
+#' Check there is no missignness in a particular vector
```
dylanhmorris, line 61:
```suggestion
#' Assert that there is no missignness in a particular vector
```

R/checkers.R
```diff
@@ -35,7 +46,8 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
   is_non_neg <- (x >= 0) | is.na(x)
   if (!all(is_non_neg)) {
     cli::cli_abort(
-      c("{.arg {arg}} has negative elements",
+      c("{.arg {arg}} has negative elements, check to ensure that values",
+        "have not been log transformed",
```
dylanhmorris, line 50:
```suggestion
      c("{.arg {arg}} has negative elements.",
```
My knee-jerk reaction is that the "log transformed" bit is too specific given the generic function name and internal logic. 

I understand that in practice this assertion will often be applied to quantities such as concentrations per mL, which are often recorded in log units but which we expect in natural units. Even then, though, that's not the only way a concentration vector could accidentally have negative values.

If you really want detailed messages for concentrations, I'd make it possible to configure the error message and add concentration-specific details to the error when it is being used to check concentrations, e.g. something like

```R
assert_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env(), additional_error_msg="")
```
and then something like
```R
cli::cli_abort(
      c("{.arg {arg}} has negative elements. ", additional_error_msg,
      "!" = "All elements must be 0 or greater",
      "i" = "Elements {.val {which(!is_non_neg)}} are negative"),
      class = "wwinference_input_data_error",
      call = call)
```

and then something like:

```R
assert_elements_non_neg(concentration_vector, additional_error_message="Note that the model expects natural scale concentration values, which must be non-negative.")

File: R/checkers.R [hunk 18]
```diff
@@ -1,13 +1,24 @@
...
+#'
+#' @param date_vector vector of dates
 #' @param max_date string indicating the maximum date in ISO8601 convention
 #' e.g. YYYY-MM-DD
 #' @param call Calling environment to be passed to the type checker
 #'
 #' @return NULL, invisibly
-check_date <- function(df, max_date, call = rlang::caller_env()) {
-  if (max(df$date) > max_date) {
+check_date_logic <- function(date_vector,
```
dylanhmorris, line 19:
```suggestion
assert_no_dates_after_max <- function(date_vector,
```

R/checkers.R
```diff
@@ -45,3 +57,268 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+#' This function  checks that the elements of a vector are
+#' not repeated.
+#'
+#' @param x the vector to check
+#' @param arg the name of the vector to check
+#' @param call Calling environment to be passed to [cli::cli_abort()] for
+#' traceback.
+#'
+#' @return NULL, invisibly
+check_for_repeat_elements <- function(x, arg = "x",
```
dylanhmorris, line 96:
```suggestion
assert_no_repeated_elements <- function(x, arg = "x",
```

R/validate.R
```diff
@@ -0,0 +1,120 @@
...
+  arg <- "ww_obs_dates"
+  check_non_missingness(ww_obs_dates, arg, call)
+  checkmate::assert_date(ww_obs_dates)
+  # check for multiple observations per day within a site and lab
+  ww_data |>
+    dplyr::group_by(site, lab) |>
+    check_for_repeat_elements()
+
+  # Sites  either need to be integers or characters, not be missing, and be
+  # non-negative
```
dylanhmorris, line 43: Why do the various ids (e.g. here site ids here) ids need to be non-negative when they are ints?

We're also implicitly relying on the fact that strings can be compared to numbers for the non-negativity check to pass. That seems potentially a bit opaque, and could have unexpected/surprising behavior. Consider for example:
```R
> "a" >= 0
[1] TRUE
> "-a" >= 0
[1] FALSE
> "a" >= Inf
[1] FALSE
> "z" >= Inf
[1] TRUE
```

Explanation: https://stackoverflow.com/a/57090193

R/checkers.R
dylanhmorris, line 1: Note that `checkmate` provides [`check_count()`](https://mllg.github.io/checkmate/reference/checkCount.html), which checks for non-negative integerish values. I think we can replace most calls to `check_elements_non_neg` and `checkmate::check_integerish()` combos with this. As for the ids, which you want to allow to be `int` or `character`, are we sure we want to enforce non-negativity on integer ids (see my comments below)? Even if so, I'd create a `check_count_or_char` that wraps `checkmate::check_count()` but also allows `char` values.

R/checkers.R
```diff
@@ -45,3 +57,268 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+                                      call = rlang::caller_env()) {
+  duplicates <- duplicated(x)
+  if (sum(duplicates) > 0) {
```
dylanhmorris, line 99:
```suggestion
  if (any(duplicates)) {
```

R/checkers.R
```diff
@@ -45,3 +57,268 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+#' @param x Object with type checking
+#' @param arg Name of the argument supplying the object
+#' @param call Calling environment to be passed to [cli::cli_abort()]
+#'
+#' @return NULL, invisibly
+check_int_or_char <- function(x, arg = "x", call = rlang::caller_env()) {
+  int_or_char_check_result <- checkmate::assert(
+    checkmate::check_integerish(x),
+    checkmate::check_character(x)
+  )
```
dylanhmorris, line 134: This will not work as written. `checkmate::assert` combines checks into an assertion, whereas you want to combine them into another check. So this will need reworking.

dylanhmorris (changes requested):

R/checkers.R
kaitejohnson: Ok can go through and change these

R/checkers.R
kaitejohnson: I like this a lot and think it makes sense, bc I can write these into the `validate` functions. Will go ahead and make this change.

R/checkers.R
kaitejohnson: My bad, I kept forgetting this.

R/validate.R
kaitejohnson: Hmm I hadn't realized that behavior, they probably don't need to be checked for non-negativity, just integer or character!

dylanhmorris: Related question: presumably any character-based indices do need to get converted to integer-based ones somewhere before they are passed to Stan. Where in the codebase does that happen, and do we have validation and/or unit testing for it?

kaitejohnson: This happens in the `get_stan_data()` function -- we probably need to add unit testing to check for the ability to do this again there!

R/checkers.R
kaitejohnson: Very odd, but I think `check_count()` only takes in a single element. Going to revert back to how we had it!

R/checkers.R
kaitejohnson: This needs to be an OR not an AND so i removed the wrapper to `checkmate::assert()`

R/checkers.R
dylanhmorris: Use `sapply`? e.g.:

```R
sapply(c(5,2,3,0.5), checkmate::check_count)
[1] "TRUE"
[2] "TRUE"
[3] "TRUE"
[4] "Must be of type 'count', not 'double'"
```

dylanhmorris: agree that I'm surprised it isn't vectorized though

kaitejohnson: I sort of hate reading code with sapply -- would prefer not to use here as this is more explicit and readable in my opinion, despite it separating the integer check from the non negativity check (which I think might be less ambigious tbh)

R/checkers.R
```diff
@@ -23,19 +35,24 @@ check_date <- function(df, max_date, call = rlang::caller_env()) {
...
-#' Check that all elements of a vector are non-negative
+#' Assert that all elements of a vector are non-negative
 #'
 #' @param x vector of arguments to check for negativity
 #' @param arg string to print the name of the element your checking
-#' @param call Calling environment to be passed to the type checker
+#' @param call Calling environment to be passed to [cli::cli_abort()] for
+#' traceback.
+#' @param add_err_msg string containing an additional error message, defauly is
+#' `NA`
```
dylanhmorris, line 45:
```suggestion
#' @param add_err_msg string containing an additional error message, default is the empty string (`""`) 
```

R/checkers.R
```diff
@@ -23,19 +35,24 @@ check_date <- function(df, max_date, call = rlang::caller_env()) {
...
-check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
+assert_elements_non_neg <- function(x, arg = "x",
+                                    call = rlang::caller_env(),
+                                    add_err_msg = "") {
   # Greater than or equal to 0 or is NA
   is_non_neg <- (x >= 0) | is.na(x)
   if (!all(is_non_neg)) {
     cli::cli_abort(
-      c("{.arg {arg}} has negative elements",
+      c("{.arg {arg}} has negative elements.",
```
dylanhmorris, line 55:
```suggestion
      c("{.arg {arg}} has negative elements.",
          add_err_msg,
```

R/checkers.R
```diff
@@ -45,3 +62,264 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+assert_no_repeated_elements <- function(x, arg = "x",
+                                        call = rlang::caller_env()) {
+  duplicates <- duplicated(x)
+  if (any(duplicates)) {
+    cli::cli_abort(
+      c("{.arg {arg}} has more than one element",
+        "i" = c(
+          "Multiple {.arg {arg}} are not currently supported.",
+          "Check that data is from a single location, and if so, provide",
+          "a single count data stream for the population in that location"
```
dylanhmorris, line 109:
```suggestion
```
This also feels like it should be an additional error message specific to the vector being checked, rather than part of the default error.

R/checkers.R
```diff
@@ -45,3 +62,264 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+#'
+#'
+#' @param x the vector to check
+#' @param arg the name of the vector to check
+#' @param call Calling environment to be passed to [cli::cli_abort()] for
+#' traceback.
+#'
+#' @return NULL, invisibly
+assert_global_pop <- function(x, arg = "x", call = rlang::caller_env()) {
+  unique_pops <- unique(x)
```
dylanhmorris, line 308: Use `assert_no_repeated_elements` here, with a custom error.

R/compile_model.R
dylanhmorris: flagging this again.

R/validate.R
```diff
@@ -0,0 +1,129 @@
...
+  # Currently, the framework only supports a single population size for
+  # an individual model fit. Therefore, check that there are not multiple
+  # "global" population sizes being passed in.
+  pop <- hosp_data |> dplyr::pull({
+    pop_size_col_name
+  })
+  arg <- "global_pop"
+  checkmate::check_count(pop)
+  assert_non_missingness(pop, arg, call)
+  assert_global_pop(pop, arg, call)
```
dylanhmorris, line 115: 1. `checkmate::check_count()` won't work here as discussed.
2. If you're going to have an `assert_global_pop` helper function, can wrap all teh hecks within it.

R/checkers.R
```diff
@@ -45,3 +62,264 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+#'
+#'
+#' @param x Object with type checking
+#' @param arg Name of the argument supplying the object
+#' @param call Calling environment to be passed to [cli::cli_abort()]
+#'
+#' @return NULL, invisibly
+assert_int_or_char <- function(x, arg = "x", call = rlang::caller_env()) {
+  count_or_char_check_result <- checkmate::check_integerish(x) ||
+    checkmate::check_character(x)
```
dylanhmorris, line 135: this won't work on failure I think as you can't `||` two strings

dylanhmorris (changes requested):
Getting very close. A few things.

R/validate.R
kaitejohnson: Got rid of the helper function

R/checkers.R
kaitejohnson: ooh good check. Hopefully would've caught this in tests. Will come up with a workaround

R/preprocessing.R
dylanhmorris: flagging this again @kaitejohnson

R/validate.R
```diff
@@ -110,17 +110,29 @@ validate_count_data <- function(hosp_data,
     pop_size_col_name
   })
   arg <- "global_pop"
-  checkmate::check_count(pop)
+  checkmate::check_integerish(pop)
+  assert_elements_non_neg(pop)
   assert_non_missingness(pop, arg, call)
-  assert_global_pop(pop, arg, call)
+  add_err_msg <- c(
```
dylanhmorris, line 116:
```suggestion
  add_err_msg <- paste0(
```

R/validate.R
```diff
@@ -110,17 +110,29 @@ validate_count_data <- function(hosp_data,
...
+    "Multiple/time-varying count catchment area populations",
```
dylanhmorris, line 117:
```suggestion
    "Multiple/time-varying count catchment area populations ",
```

R/validate.R
```diff
@@ -110,17 +110,29 @@ validate_count_data <- function(hosp_data,
...
+  assert_no_repeated_elements(pop, arg, call, add_err_msg)
 
 
   # Date column should be of date type, for count data, there should only
   # be on observation per day
   count_dates <- hosp_data$date
   arg <- "count_obs_dates"
   checkmate::assert_date(count_dates)
-  assert_no_repeated_elements(count_dates, arg, call)
+  add_err_msg <- c(
```
dylanhmorris, line 130:
```suggestion
  add_err_msg <- paste0(
```

R/validate.R
```diff
@@ -110,17 +110,29 @@ validate_count_data <- function(hosp_data,
...
+    "Check that data is from a single location, and if so,",
```
dylanhmorris, line 131:
```suggestion
    "Check that data is from a single location, and if so, ",
```

R/validate.R
```diff
@@ -0,0 +1,141 @@
...
+                                  conc_col_name,
+                                  lod_col_name,
+                                  call = rlang::caller_env()) {
+  ww_conc <- ww_data |> dplyr::pull({
+    conc_col_name
+  })
+  arg <- conc_col_name
+  assert_non_missingness(ww_conc, arg, call)
+  assert_elements_non_neg(ww_conc, arg, call,
+    add_err_msg = c(
```
dylanhmorris, line 22:
```suggestion
    add_err_msg = paste0(
```

R/validate.R
```diff
@@ -0,0 +1,141 @@
...
+  )
+  checkmate::assert_vector(ww_conc)
+
+  ww_lod <- ww_data |> dplyr::pull({
+    lod_col_name
+  })
+  arg <- "lod_col_name"
+  assert_non_missingness(ww_lod, arg, )
+  assert_elements_non_neg(ww_lod, arg, call,
+    add_err_msg = c(
```
dylanhmorris, line 36:
```suggestion
    add_err_msg = paste0(
```

R/validate.R
```diff
@@ -0,0 +1,141 @@
...
+      "Note that the model expects natural ",
+      "scale LOD values, which must be",
```
dylanhmorris, line 38:
```suggestion
      "scale LOD values, which must be ",
```

R/validate.R
```diff
@@ -0,0 +1,141 @@
...
+
+  # Wastewater date column should be of date type!
+  ww_obs_dates <- ww_data$date
+  arg <- "ww_obs_dates"
+  assert_non_missingness(ww_obs_dates, arg, call)
+  checkmate::assert_date(ww_obs_dates)
+  # check for multiple observations per day within a site and lab
+  ww_data |>
+    dplyr::group_by(site, lab) |>
+    assert_no_repeated_elements()
```
dylanhmorris, line 52: I am fairly sure this will not work as intended. `assert_no_repeated_elements()` expects a vector, not a data frame. `duplicated()` will work on a data frame or `tibble`, but it identifies cases where _all columns are duplicated_, which is not what you want here, and it also will not automatically be applied over `dplyr::group_by()` groups. Example:

```R
> df <- tibble::tibble(x=c(1,1,1,2,2), y=c(1,1:4), z=c(1,1,1,1,1))
> df
# A tibble: 5 × 3
      x     y     z
  <dbl> <dbl> <dbl>
1     1     1     1
2     1     1     1
3     1     2     1
4     2     3     1
5     2     4     1
>
> df |> duplicated()
[1] FALSE  TRUE FALSE FALSE FALSE
>
> df |> dplyr::group_by(x,y) |> duplicated()
[1] FALSE  TRUE FALSE FALSE FALSE
>
> df |> dplyr::group_by(x,y) |> dplyr::summarise(dup = any(duplicated(z)))
`summarise()` has grouped output by 'x'. You can override using the `.groups`
argument.
# A tibble: 4 × 3
# Groups:   x [2]
      x     y dup
  <dbl> <dbl> <lgl>
1     1     1 TRUE
2     1     2 FALSE
3     2     3 FALSE
4     2     4 FALSE

Comment (dylanhmorris) [reply R/validate.R:52]: For this reason, it will be important to implement tests for these checkers, ideally checking at least one expected pass and one expected failure for each.

File: R/validate.R [hunk 37]
```diff
@@ -0,0 +1,141 @@
...
+  add_err_msg <- c(
+    "Check that data is from a single location, and if so,",
+    "ensure that there are not multiple count data streams"
+  )
+  assert_no_repeated_elements(count_dates, arg, call, add_err_msg)
+
+
+
+
+
```
dylanhmorris, line 139: Extra whitespace

R/validate.R
```diff
@@ -0,0 +1,141 @@
...
+    "Multiple/time-varying count catchment area populations",
+    "are not currently supported. Check that data is from a ",
+    "single location, and if so, consider replacing with an ",
+    "average population size over the inference period"
+  )
+  assert_single_value(pop, arg, call, add_err_msg)
+
+
+  # Date column should be of date type, for count data, there should only
+  # be on observation per day
```
dylanhmorris, line 126:
```suggestion
  # be one observation per day
```

dylanhmorris (changes requested):

R/validate.R
kaitejohnson: Also I got confused and these should be all the same :( (so all repeats)

R/checkers.R
```diff
@@ -45,3 +62,258 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+#' traceback.
+#' @param add_err_msg string containing an additional error message,
+#' default is the empty string (`""`)
+#'
+#' @return NULL, invisibly
+assert_no_repeated_elements <- function(x, arg = "x",
+                                        call = rlang::caller_env(),
+                                        add_err_msg = "") {
+  duplicates <- duplicated(x)
+  if (any(duplicates)) {
```
damonbayer, line 110: You probably want `anyDuplicated`

R/checkers.R
```diff
@@ -45,3 +62,258 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+#' @param call Calling environment to be passed to [cli::cli_abort()]
+#'
+#' @return NULL, invisibly
+assert_int_or_char <- function(x, arg = "x", call = rlang::caller_env()) {
+  # Check if its a character, if it is, check passes. If not,
+  # check if its an integer
+  int_or_char_check_result <- is.character(x)
+  if (!int_or_char_check_result) {
+    int_or_char_check_result <- checkmate::check_integerish(x)
+  }
```
dylanhmorris, line 142:
```suggestion
is_integerish <- checkmate::check_integerish(x)
is_character <- checkmate::check_character(x)
is_int_or_char <- isTRUE(is_integerish) || isTRUE(is_character)
```

R/checkers.R
```diff
@@ -45,3 +62,258 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+
+  if (!int_or_char_check_result) {
```
dylanhmorris, line 144:
```suggestion
  if (!is_int_or_char) {
```

R/checkers.R
```diff
@@ -45,3 +62,258 @@ check_elements_non_neg <- function(x, arg = "x", call = rlang::caller_env()) {
...
+    throw_type_error(
+      object = x,
+      arg_name = arg,
+      expected_type = "integer or character",
+      call = call
+    )
+  }
```
dylanhmorris, line 151: In the future we can leverage the `checkmate::check_` string output to give a more verbose failure message, but that's a separate PR

dylanhmorris (changes requested):
Minor typo fixes and suggestions, then should be ready.

I do think that in the future you will want validation by group, but that can be a separate PR.

dylanhmorris (approved):
LGTM!

Issue #28, opened 2024-07-17
Stan spatial functions.


cbernalz:
## Goal
We want to create STAN functions that are similar to the R functions for the spatial components of the model.

## Context
The following model will be added as STAN functions :
Let $k$ represent a site, from sites $1$ to $n_k$.  Let $R_k^u(t)$ be the unadjusted reproduction number at time $t$ at site $k$.  So lets look at the combined "state" $R^u(t)$ that contains an AR(1) correlation structure on the first differences that is currently being used.  So 

$$
\log R^u(t) \sim \text{Normal}(\log R^u(t - 1) + \beta(\log R^u(t - 1) - \log R^u(t - 2)), \sigma_r),
$$

or

$$
\log R^u(t) = \log R^u(t - 1) + \beta(\log R^u(t - 1) - \log R^u(t - 2)) + \sigma_r\text{Normal}(0,1).
$$

Now let $\delta_k(t)$ be the time-varying subpopulation effect on $\log R^u(t)$.  So we can look at the vector of all $\log R_k^u(t)$'s, so

$$
\begin{pmatrix}
        \log R_1^u(t)\\
	\vdots\\
	\log R_{n_{k}}^u(t)\\
\end{pmatrix} = \log R^u(t)
\begin{pmatrix}
	1\\
	\vdots\\
	1\\
\end{pmatrix} + \vec{\delta}(t).
$$

Where $\vec{\delta}(t) = \left(\delta_1(t),...,\delta_{n_k}(t)\right)^{T}$.  We also have that $\vec{\delta}(t) = \varphi_{R(t)}\vec{\delta}(t - 1) + \vec{\epsilon}$, where $\vec{\epsilon}$ $\sim$ $\text{Normal}_{n_k}(0,\mathbf{\Sigma\_{\epsilon}}).$  

This adds a spatial correlation and also the current AR(1) temporal correlation to this $\vec{\delta}(\cdot)$ random variable.  Like before, we have that $0 < \varphi_{R(t)} < 1$.

For $\mathbf{\Sigma_{\epsilon}}$  $=$  $\sigma_{\epsilon}^2\mathbf{\Omega}$, where $\Omega_{ij} \in \mathbf{\Omega}$ are build from the correlation functions.  These include an exponential decay, Matern, spherical, rational quadratic, or, to return to the current model framework, independence correlation functions.  

For simplicity, we will first assume an distance matrix is provided with the data or locations of sites where we can attain an distance matrix.

[...]

PR #39, opened 2024-07-30
2024-07-30 update : added exp decay func test.

cbernalz:


tests/testthat/setup.R
```diff
@@ -21,6 +21,16 @@ compiled_site_inf_model <- cmdstanr::cmdstan_model(
   dir = tempdir()
 )
 
+spatial_fxns <- cmdstanr::cmdstan_model(
+  stan_file = system.file(
+    "stan", "functions", "spatial_functions.stan",
+    package = "wwinference"
+  ),
+  compile = TRUE,
```
kaitejohnson, line 29: Suggest adding the `dir = tempdir()`  as in the compilation of the `wwinference.stan` model, but otherwise this looks right!

tests/testthat/test_exponential_decay_corr_func.R
```diff
@@ -0,0 +1,35 @@
...
+    exp_decay_corr_func_stan <-
+      space_model_fxns$functions$exponential_decay_corr_func(
+        corr_func_params$dist_matrix,
+        corr_func_params$phi,
+        corr_func_params$l
+      )
+    exp_decay_corr_func_r <- exponential_decay_corr_func_r(
+      corr_func_params
+    )
+
```
kaitejohnson, line 28: This looks good to me! Maybe just add some comments about what is being returned by both of these functions.

tests/testthat/test_spatial_deviation_noise_matrix_rng.R
```diff
@@ -0,0 +1,67 @@
...
+    )
+    sqrt_sigma_matrix <- expm::sqrtm(
+      sigma_matrix
+    )
+    r_noise_matrix <- apply(
+      t(r_non_center_noise_matrix),
+      2,
+      function(x) sqrt_sigma_matrix %*% x
+    )
+
```
kaitejohnson, line 54: Why do you need the column wise and apply function ? 
Wouldn't it just be:
`r_noise_matrix <- sqrt_sigma_matrix %*% t(r_non_center_noise_matrix)`

tests/testthat/test_spatial_deviation_noise_matrix_rng.R
```diff
@@ -0,0 +1,67 @@
...
+    # comparing with cramer test
+    cramer_p_value <- cramer.test(
+      stan_noise_matrix,
+      r_noise_matrix
+    )$p.value
+
+
+    testthat::expect_gt(
+      cramer_p_value,
```
kaitejohnson, line 63: Can you explain to me like I am five how the cramer test works? A p-value of > 0.05 means that the two matrices are not statistically significantly different?

I would think you would want something like repeated across many simulations, generate the noise matrices both ways, and then compare the distributions of matrix elements to show they are similar enough/have the same statistical properties (so perhaps the same trace, the same determinant, etc)?

cbernalz:
Before official review, @dylanhmorris and / or @kaitejohnson, how can we make a test for the auxiliary site functions or are we fine not having one for those?

tests/testthat/test_spatial_deviation_noise_matrix_rng.R
cbernalz: @kaitejohnson we need column wise, because the columns are the random vectors that we sample, then using the transformation `sqrt_sigma_matrix %*% column` gets us to the $\text{MVN}(0,\Sigma\_{\epsilon})$ for that new transformed vector.  The entire noise matrix is just `n_times` realizations of the $\text{MVN}(0,I)$.

tests/testthat/test_spatial_deviation_noise_matrix_rng.R
cbernalz: @kaitejohnson the cramer test is a way to compare two multivariate samples, where the null hypothesis is that they are coming from the same distribution.  So, we "want" a high p-value, which gives us more confidence that these two multivariate samples come from the same distribution.  The test statistic it uses is in the paper I added in the spatial folder on sharepoint titled, `On a new multivariate two-sample test`, but this test statistic looks at the euclidean distance of two sampled vectors, with this I believe they make it easier to find the distribution of the test statistic to give us a way to get the p-value.

cbernalz: @kaitejohnson I think it would be a good idea to do this across many simulations instead of just one seed, like I currently have it.  But I don't think looking at the trace or determinant would work since this noise matrix is `n_site x n_time`, which isn't typically a square matrix.  And if I am not mistaken, trace and determinant are only for square matrices.

kaitejohnson: ^You are definitely right about that :) Although for the test you could make them square matrices but probably best not to in order to keep it closest to the real situation

I might see if @athowes has thoughts here as this is somewhat related to unit testing when sampling from a target distribution. I just am sort of wary of relying on a p_value > 0.05 as a robust test to tell us that two draws are in fact coming from the same distribution, but yeah my brain breaks when we start to think about doing this across samples from matrices

kaitejohnson:
> Before official review, @dylanhmorris and / or @kaitejohnson, how can we make a test for the auxiliary site functions or are we fine not having one for those?

I think it would be good to have a test that shows that variance in the deviations in the auxiliary site are the same as the variance in the spatially correlated sites (so related to this week's discussion and https://github.com/cdcent/cfa-spatial-ww-internal/issues/4)

tests/testthat/test_exponential_decay_corr_func.R
```diff
@@ -0,0 +1,38 @@
...
+        corr_func_params$phi,
+        corr_func_params$l
+      )
+    exp_decay_corr_func_r <- exponential_decay_corr_func_r(
+      corr_func_params
+    )
+    # both functions return correlation matrix using an
+    # exponential decay correlation function
+
+    testthat::expect_equal(
```
dylanhmorris, line 32: You may want `expect_almost_equal()` here to make the test robust to potential minor differences in float arithmetic

tests/testthat/test_independence_corr_func.R
```diff
@@ -0,0 +1,27 @@
...
+      space_model_fxns$functions$independence_corr_func(
+        corr_func_params$num_sites
+      )
+    ind_corr_func_r <- independence_corr_func(
+      corr_func_params
+    )
+    # both functions return correlation matrix using an
+    # independent correlation function
+
+    testthat::expect_equal(
```
dylanhmorris, line 21: Again, may want `expect_almost_equal` here

tests/testthat/test_exponential_decay_corr_func.R
```diff
@@ -0,0 +1,38 @@
+test_that(
+  "Test Exponential Decay Corr Func in Stan agrees with R",
+  {
+    model <- compiled_site_inf_model
+    space_model_fxns <- spatial_fxns
+
+    corr_func_params <- list(
```
dylanhmorris, line 7: set the seed here for reproducibility? also, I'd suggest using `withr::with_seed` rather than just calling `set.seed` within the individual tests https://withr.r-lib.org/reference/with_seed.html

dylanhmorris, line 7: would be nice to run this test multiple times with multiple random `corr_func_params` parameter sets, including different `phi` and `l` values. Definitely worth checking with `l=1`

tests/testthat/test_independence_corr_func.R
```diff
@@ -0,0 +1,27 @@
+test_that(
+  "Test Independence Corr Func in Stan agrees with R",
+  {
+    model <- compiled_site_inf_model
+    space_model_fxns <- spatial_fxns
+
+    corr_func_params <- list(
```
dylanhmorris, line 7: - try for a few different site counts including 1?

dylanhmorris (commented):
A few minor comments / questions / suggestions as I go @cbernalz.

Also, are the functions set up to error with pad inputs, e.g. you try to use `num_sites` that is not a positive integer? 

- If so, we can and should add an `expect_error` test to confirm this occurs
- If not, we should make them do so, but that can be a separate issue / PR, and we can add those tests there.

tests/testthat/test_exponential_decay_corr_func.R
cbernalz: @dylanhmorris which library is that in, it doesn't seem to be in the testthat library.  expect_equal() does have a tolerance option if that also works.

dylanhmorris: My mistake, thanks @cbernalz. I forgot that in `testthat`, [`expect_equal()` is the name of the float-tolerant expectation and `expect_identical()` is the name of the stricter expectation](https://testthat.r-lib.org/reference/equality-expectations.html).

tests/testthat/test_independence_corr_func.R
dylanhmorris: Ignore this, see https://github.com/CDCgov/ww-inference-model/pull/39#discussion_r1700816804

tests/testthat/test_spatial_deviation_noise_matrix_rng.R
athowes: Don't fully understand context here but I have also used relying on p-value greater than 0.01 (rather than 0.05 but same idea) as a test. Summary of discussion was that moving the p-value closer to 0 and increasing sample size will reduce false negative rate.

And about the matrices, the point is that you have some distribution on spatio-temporal random effects, and conceptualise that as a distribution on matrices where one dimension is space and one is time? Seems possible. I wonder if there are any simpler tests on decompositions of that, but maybe you have those too already.

athowes: From `?cramer.test` it looks like some kernel-based distance test. If it's a multivariate test, do you know that it's valid for matrices? I guess what you would be doing is putting the matrix into a vector (maybe all tests for vectors can be applied to matrices by doing that)

cbernalz: @athowes this does bring an interesting point to using / trusting the cramer test.  I believe one of the assumptions is that observations of each sample are independent of each other.  In our case we know they are not, since we are looking at AR(1) on time differences, so I don't believe we can even use this test here.  But this test is valid for matrices that contain random vectors of some distribution, but after looking at the documentation a bit more, I need to switch how I am passing in the matrix into `cramer.test`.  

Here is the paper I looked at to see if I could use this test, just in case you don't have access to the spatial folder in sharepoint :
[On a new multivariate two-sample test.pdf](https://github.com/user-attachments/files/16472378/On.a.new.multivariate.two-sample.test.pdf)

@athowes and / or @kaitejohnson do you know of any multivariate tests that don't have the this independence assumption or should we not be concerned about it?

kaitejohnson: I think an alternative might be to apply a test to just a single time point (so the resulting vector), since the point of this test is to ensure that the non-centered parameterization via the cholesky decomposition is equivalent to the original centered parameterization? 

That way, we could just generate $n$ vectors from each MVN and use the KS test for comparing the vectors? Does that make sense? Perhaps am over simplifying

kaitejohnson:
> > Before official review, @dylanhmorris and / or @kaitejohnson, how can we make a test for the auxiliary site functions or are we fine not having one for those?
> 
> I think it would be good to have a test that shows that variance in the deviations in the auxiliary site are the same as the variance in the spatially correlated sites (so related to this week's discussion and [cdcent/cfa-spatial-ww-internal#4](https://github.com/cdcent/cfa-spatial-ww-internal/issues/4))

NVM forgot we were implementing with the additional parameter of a scaling factor

tests/testthat/test_spatial_deviation_noise_matrix_rng.R
cbernalz: After a conversation with Kaitlyn, non-independence among random vectors will not be a problem since we are not looking at the $\vec{\delta}(t)$ term from the model in https://github.com/CDCgov/ww-inference-model/issues/28#issue-2414103263, we are looking at the noise term, or the $\vec{\epsilon}$ term.  So I believe it is fine to do the cramer test for this case, I just need to fix how we pass in the matrix of random vectors into the `cramer.test` function as it accepts a matrix with rows as random vectors and I currently have it set up as a matrix with columns as random vectors.

DESCRIPTION
```diff
@@ -83,7 +83,10 @@ Imports:
     scales,
     ggplot2,
     posterior,
-    MASS
+    MASS,
+    cramer,
+    expm,
+    rstan
```
dylanhmorris, line 89: If `rstan` is just used in the `scratch` or even in the `vignettes`, as here, it can be in `Suggests`. But any particular reason why you used `rstan` versus `cmdstanr` for those `Rmd` examples?

cbernalz: I did not know that cmdstanr was an alternate to rstan.  I can switch them.

kaitejohnson: From my memory, the syntax for writing an example stan file directly in R is nicer in `rstan` ! Let me see if I can find the syntax of cmdstanr. You can't just swap them in and out, I don't think

kaitejohnson: I can't seem to find it, but I am 90% positive its more annoying than just putting it in quotations...

tests/testthat/test_spatial_deviation_noise_matrix_rng.R
```diff
@@ -0,0 +1,73 @@
...
+      ),
+      phi = 25,
+      l = 1
+    )
+    omega_matrix <- space_model_fxns$exponential_decay_corr_func(
+      corr_fun_params$dist_matrix,
+      corr_fun_params$phi,
+      corr_fun_params$l
+    )
+    sigma_matrix <- sqrt(0.02) * omega_matrix
```
kaitejohnson, line 27: I would define `sigma_eps = sqrt(0.02)` for readability. 

Also, just confirming that this doesn't include any normalizing by the maximum distance or ensuring that the trace of $$\Sigma$$ is equal to the number of sites?

tests/testthat/test_spatial_rt_process.R
```diff
@@ -0,0 +1,56 @@
...
+    )
+    # these functions creates the spatial Rt for the site level given
+    # the correlation or covariance structure.
+    # We use cramer test for multi-variate random vectors
+
+    cramer_p_value <- cramer.test(
+      log_site_rt_r,
+      log_site_rt_stan
+    )$p.value
+
```
kaitejohnson, line 50: I think like in `test_spatial_deviation_noise_matrix` we might want to repeat this test `n_times`

tests/testthat/test_aux_site_functions.R
```diff
@@ -0,0 +1,63 @@
...
+      state_deviation_t_i <- 0
+      for (t_i in 1:n_times) {
+        state_deviation_t_i <- state_deviation_ar_coeff * state_deviation_t_i +
+          state_deviation_noise_vec[t_i]
+        r_log_aux_site_rt[t_i] <- log_state_rt[t_i] + state_deviation_t_i
+      }
+      r_log_aux_site_rt <- as.vector(r_log_aux_site_rt)
+
+      # comparing with ks test
+      ks_p_value <- ks.test(
```
kaitejohnson, line 47: The other way to do this that I think might be slightly more idiomatic (but will defer to @dylanhmorris and @athowes on that as this is all fairly new to me as well), is to instead take a ks test of the distributions across all of the `n_tests` since I believe here they are independent. 

And then you would just need to use a single p-value threshold. I think this is what is meant by running a lot of simulations to ensure you're not likely to have the tests fail by chance/randomness.

scratch/stan_chol_exmpl.R
```diff
@@ -129,13 +133,19 @@ sigma_matrix_true <- sigma_eps_true^2 * exponential_decay_corr_func_r(list(
...
+epsilon <- matrix(data = 0, nrow = n_sites, ncol = n_times)
 obs_data <- matrix(data = 0, nrow = n_sites, ncol = n_times)
 for (i in 1:n_times) {
-  obs_data[, i] <- t(mvrnorm(
+  epsilon[, i] <- t(mvrnorm(
     n = 1,
     mu = rep(0, n_sites),
     Sigma = sigma_matrix_true
   ))
+  obs_data[, i] <- t(mvrnorm(
```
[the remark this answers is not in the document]
kaitejohnson: It helps me a lot when writing these examples to use Rmds because I can put the Latex math in right above and around what I am coding, which is a good reminder of how the centered and non-centered models differ mathematically.

tests/testthat/test_spatial_rt_process.R
kaitejohnson: Can be a separate issue though

scratch/stan_chol_exmpl.R
```diff
@@ -0,0 +1,200 @@
...
+  int<lower=1> n_times;
+  matrix[n_sites,n_sites] dist_matrix;
+  real<lower=0> l;
+  matrix[n_sites,n_times] obs_data;
+}
+
+parameters {
+  real<lower=0> phi;
+  real<lower=0> sigma_eps;
+  matrix[n_sites, n_times] epsilon;
```
kaitejohnson, line 79: In the centered parameterization, we don't have to estimate an expected value of epsilon

kaitejohnson, line 79:
```suggestion
```

scratch/stan_chol_exmpl.R
```diff
@@ -0,0 +1,200 @@
...
+  Sigma = sigma_eps^2 * exponential_decay_corr_func(dist_matrix, phi, l);
+}
+
+model {
+  //Priors
+  phi ~ uniform(0, 50);
+  sigma_eps ~ gamma(sqrt(0.1),1);
+  for (i in 1:n_times) {
+    epsilon[,i] ~ multi_normal(rep_vector(0, n_sites), Sigma);
+  }
```
kaitejohnson, line 94:
```suggestion
```

scratch/stan_chol_exmpl.R
```diff
@@ -0,0 +1,200 @@
...
+
+  // Likelihood
+  for (i in 1:n_times) {
+    obs_data[,i] ~ normal(epsilon[,i], 1);
```
kaitejohnson, line 98:
```suggestion
    obs_data[,i] ~ multi_normal(0, Sigma);
```

kaitejohnson (changes requested):
I think this is super close @cbernalz just a few questions/suggested changes. They key one being the NCP stan_chol_exmpl -- my understanding of the example is that the data generating model is $\epsilon \sim MVN(0, \sigma)$ and you want to test and compare the centered and non-centered implementation of fitting to the same observed data $\epsilon$. 

The confusing thing is that in the non- centered parameterization, we generate expected values of $\epsilon$, but in the centered parameterization, this is just our observation model, so we don't have to estimate an epsilon parameter. This will be less confusing in the stan model when the expected values from the MVN that we are doing the non-centered parameterization on feed into another function rather than being the observations.

scratch/stan_chol_exmpl.R
```diff
@@ -0,0 +1,200 @@
+library(tidyverse)
+library(rstan)
+
+
+
```
kaitejohnson, line 5: This is where I would add (in comments since its a script) the model youre fitting to!

kaitejohnson (commented):
Can ignore my previous review, didn't understand the model we were using. My bad! Only question is about the p-values across the many simulations and then this looks good to me

DESCRIPTION
dylanhmorris: `cmdstanr` provides a [`write_stan_file`](https://mc-stan.org/cmdstanr/reference/write_stan_file.html) function for this use case.

dylanhmorris (commented):
Tests look good. Main remaining questions are about the `Rmd` files, as I do think we should avoid the `rstan` additional dependency if possible. @cbernalz @kaitejohnson

kaitejohnson:
> Tests look good. Main remaining questions are about the `Rmd` files, as I do think we should avoid the `rstan` additional dependency if possible. @cbernalz @kaitejohnson

That sounds good. I think we should just replace with that, I didn't realize that function existed.

kaitejohnson (dismissed):
Pending CI passes, this looks good to me

dylanhmorris:
> Pending CI passes, this looks good to me

I thought we were going to replace `rstan` with `cmdstanr` before landing this?

cbernalz:
> > Pending CI passes, this looks good to me
> 
> I thought we were going to replace `rstan` with `cmdstanr` before landing this?

@dylanhmorris I thought I replaced all of them.  Can you reference where this is still the case?

dylanhmorris (approved):
LGTM

PR #40, opened 2024-07-31
Quick fix to R(t) plots and markdown rendering

kaitejohnson:
Caught a few things when looking at the rendered getting started vignette on the pkgdown site.


dylanhmorris (approved):
LGTM

PR #47, opened 2024-08-02
Slight tweak to the centered code

kaitejohnson:
I now realize why this was very confusing! Because when you do the non-centered parameterization, you generate expectations of the observed data, and these expectations become the mean of your observation model (and I chose normally distributed observation noise here, though I am not sure that is necessarily best practice). 

For the centered paramaterization, you want to write the model in the exact way you are simulating the data which is that:

$$\epsilon_t \sim MVN(0, \Sigma)$$

Where $\Sigma$ is just a function of the data (distance matrix) and the correlation function parameters (`phi` and `sigma_eps`). In this case, you do not have to parameterize the "noise matrix" or the realized draws from the iid normals, because they will go directly into your observation model as part of the likelihood function.

### DIFF
--- a/scratch/stan_chol_exmpl.R
+++ b/scratch/stan_chol_exmpl.R
@@ -129,17 +129,15 @@ sigma_matrix_true <- sigma_eps_true^2 * exponential_decay_corr_func_r(list(
   phi = phi_true,
   l = l
 ))
-epsilon_true <- t(mvrnorm(
-  n = n_times,
-  mu = rep(0, n_sites),
-  Sigma = sigma_matrix_true
-))
+# Generate samples from a MVM with mean 0 and covariance matrix sigma
+# epsilon_t ~ MVN(0, sigma_matrix) # nolint
+# for each time point. These are your observations!
 obs_data <- matrix(data = 0, nrow = n_sites, ncol = n_times)
 for (i in 1:n_times) {
   obs_data[, i] <- t(mvrnorm(
     n = 1,
-    mu = epsilon_true[, i],
-    Sigma = diag(1, n_sites)
+    mu = rep(0, n_sites),
+    Sigma = sigma_matrix_true
   ))
 }
 
--- a/scratch/stan_chol_exmpl.R
+++ b/scratch/stan_chol_exmpl.R
@@ -24,6 +24,7 @@ parameters {
   matrix[n_sites,n_times] alpha;
   real<lower=0> phi;
   real<lower=0> sigma_eps;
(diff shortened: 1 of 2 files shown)


scratch/stan_chol_exmpl.R
```diff
@@ -38,13 +39,20 @@ transformed parameters {
...
+  // here,
 
   // Likelihood
+  // Assumes normally distributed observation error.
+  // obs_data are your vectors of draws the MVN at each time point (column
+  // bound to produce a matrix). The epsilon matrix are the
+  // mean of the expected realizations of the MVN draws.
   for (i in 1:n_times) {
-    obs_data[,i] ~ multi_normal(epsilon[,i], identity_matrix(n_sites));
+    obs_data[,i] ~ normal(epsilon[,i], sigma_obs);
```
cbernalz, line 55: This all looks good, but I am a bit confused as to how `sigma_obs` matches with the centered stan model.  Why isn't this term on the centered stan model?

cbernalz (approved):

scratch/stan_chol_exmpl.R
kaitejohnson: The term isn't on the centered stan model because the values drawn from the MVN are being evaluated directly with the call `obs_data ~multinormal(0, Sigma)`

Whereas in the non-centered parameterization, we generate the expected values `epsilon` from the MVN using the cholesky decomposition transformation, and then we need to evaluate the observed data compared to the expected values `epsilon`, and this requires specifying a probability density (I chose normally distributed error but it could be anything, the point is just that we have to evaluate the log density of the observed data vs the expectation). 

I am going to post a simple example of a NCP and try to walk through it!  Also quite possible that I am over complicating this/made an error in this example, but I can't think of a way to do it without adding an additional `sigma_obs` term of some sorts. 

In our `wwinference` model, we already have an observation model and the expected values of the MVN get passed into another function and further transformed, so we don't have to add an extra parameter. It's just that in this simple example the MVN is our observation model.

cbernalz: Yea, it could have just been a bad example on my part.  But I looked over the files you added and they look good, noncentered vs centered in stan makes a bit more sense to me now.

kaitejohnson: No no was not a bad example on your part, I think it makes sense to default to the simplest model and I probably suggested it! I am adding one more example where the thing you non-center is a latent variable and am hoping it will make more sense!

cbernalz (approved):

Issue #51, opened 2024-08-06
Standardize to use `count_data` in all functions


kaitejohnson:
https://github.com/CDCgov/ww-inference-model/pull/37#discussion_r1705914175

dylanhmorris:
Has this been done?

kaitejohnson:
Yes

Issue #49, opened 2024-08-06
Standardize and stabilize outputs of `wwinference()` wrapper function 


kaitejohnson:
## Goal
Before a first release (checklist [here](https://github.com/CDCgov/ww-inference-model/issues/6)), we will want to make sure that the main model fitting wrapper function, `wwinference()` returns a stable set of contents. Right now, we are returning a post-processed `draws` dataframe + mappings to sites and labs, dates, and maps to the subpopulations.  We are also doing a diagnostic check internally within the wrapper function, while also showing the user how to do this from the stan fit object separately in the vignette.

I think it would be best that we don't mix post-processing and model fit at all, and instead write the `wwinference()` function such that it returns a nested list, `ww_fit` containing `raw_stan_fit_obj`, `input_data`, and `stan_data_args`. Only the first two will be used in downstream functions, but it might be helpful, similar to the passing around the input data, to pass around the inputs to the stan model (e.g. the priors) for comparing outputs with different priors. 

From this, we would overload downstream functions to take in the `ww_fit` object and generate the draws dataframe, run the diagnostics, plot the results, and (eventually) evaluate the results. The main change from the current workflow right now would be to adjust `get_draws_df()` to take in the `ww_fit` object and remove the internal calls to `get_draws_df()` and `get_model_diagnostic_flags()`

The goal of all of this would be to 1 modularize the codebase so each function performs one step in the workflow, and 2 put the onus on us to demonstrate the workflow of the package step-by-step and the user to perform each action on the fitted model object. 

## Requirements
- [ ] output the following elements from `wwinference()`: `raw_stan_fit_obj`, `input_data`  and `stan_data_args`. All will be lists. Optionally, we could separate the `ww_data` and `hosp_data` and pass back both as dataframes
- [ ] rewrite `get_draws_df()` (perhaps rename to `extract_posterior_draws()`) to take in the `ww_fit` object as an output and return something that looks like `draws_df()` currently. Allow user to pass in args to only include certain variables among a set of allowed variables.
- [ ] revise vignette to demonstrate each of these steps
- [ ] add a mermaid diagram describing how the functions interact
- [ ] all revised functions (fitting wrapper functions and downstream functions) should be tested

@seabbs @gvegayon @dylanhmorris Curious any other thoughts here

gvegayon:
Without fully understanding the details, I think the key here is leveraging S3 methods:

[...]

Issue #21, opened 2024-07-12
Make the small added value for `log_conc` configurable


dylanhmorris:
> Would be good to make the small added on value (here `1e-8`) configurable, but that can be a separate PR.

_Originally posted by @dylanhmorris in https://github.com/CDCgov/ww-inference-model/pull/11#discussion_r1676084988_

Issue #17, opened 2024-07-10
Avoid the `assign(param_names[i])` construct


dylanhmorris:
> Sure, how would you suggest fixing it? I agree its clunky. Could just unlist a list? Would prefer to not have to always call things by params$parameter_name, unless that is the best practice.

_Originally posted by @kaitejohnson in https://github.com/CDCgov/ww-inference-model/pull/5#discussion_r1671473868_

dylanhmorris:
This is complete as of #29

PR #54, opened 2024-08-07
Testing for key functions

### LINKED ISSUE #27: Add testing for key functions
## Goal
We created a bunch of new functions, and we should add a testing suite for them. The functions that need to be tested include:
`preprocess_ww_data` `preprocess_hosp_data` `flag_ww_outliers`, `indicate_ww_exclusions` in `preprocessing.R` `get_draws_df` in `get_draws_df.R` and the rest of the functions in `utils.R`.

## Requirements
- [ ] additional tests in `tests` folder for each of these functions

### LINKED ISSUE #48: Tests for autoindexing
> This happens in the `get_stan_data()` function -- we probably need to add unit testing to check for the ability to do this again there!

_Originally posted by @kaitejohnson in https://github.com/CDCgov/ww-inference-model/pull/37#discussion_r1704382679_

kaitejohnson:
Goal is just to have some amount of unit tests of the functions in place. In this case, because I am planning on making some significant revisions in #49 , this PR is focused on testing the expected behavior of the pre-processing, the input data validation, and then the immediate validation/tests of the inputs being passed to stan. It is not intending to be comprehensive. I have added to the description in #49 to specify that that PR requires that all those functions be tested.

 This will close #27, but also some scope creep/clean up will close #48, #17, #51, and #21.


R/checkers.R
dylanhmorris, line 1: Should've caught this before; apologies. The error message for `assert_single_value` should be something like `"{.arg {arg}} should have a single unique value; it has more than one"` rather than "{.arg {arg}} has more than one element"

R/checkers.R
```diff
@@ -317,3 +326,173 @@ assert_single_value <- function(x, arg = "x",
...
+      ),
+      call = call,
+      class = "wwinference_input_data_error"
+    )
+  }
+  invisible()
+}
+
+#' Assert that the vector of dates contains the duration of the specified
+#' calibration time
```
dylanhmorris, line 393:
```suggestion
#' Assert that the vector of dates spans at least the specified
#' calibration time
```

R/checkers.R
```diff
@@ -317,3 +326,173 @@ assert_single_value <- function(x, arg = "x",
...
+#' @param add_err_msg add_err_msg string containing an additional error message,
+#' default is the empty string (`""`)
+#'
+#' @return NULL invisible
+assert_sufficient_days_of_data <- function(date_vector,
+                                           calibration_time,
+                                           call = rlang::caller_env(),
+                                           add_err_msg = "") {
+  # check that you have sufficient count data for the calibration time
+  min_date <- max(date_vector,
```
dylanhmorris, line 409:
```suggestion
  calibration_start <- max(date_vector,
```

R/checkers.R
```diff
@@ -317,3 +326,173 @@ assert_single_value <- function(x, arg = "x",
...
+    na.rm = TRUE
+  ) - lubridate::days(calibration_time) + 1
+  check_sufficient_data <- min(date_vector, na.rm = TRUE) <= min_date
```
dylanhmorris, line 412:
```suggestion
  check_sufficient_data <- min(date_vector, na.rm = TRUE) <= calibration_start
```

R/checkers.R
```diff
@@ -317,3 +326,173 @@ assert_single_value <- function(x, arg = "x",
...
+#' @param date_vector the vector of dates to check, must be of Date type
+#' @param calibration_time integer indicating the number of days that
+#' the dates must span
+#' @param call Calling environment to be passed to [cli::cli_abort()] for
+#' traceback.
+#' @param add_err_msg add_err_msg string containing an additional error message,
+#' default is the empty string (`""`)
+#'
+#' @return NULL invisible
+assert_sufficient_days_of_data <- function(date_vector,
```
dylanhmorris, line 404: Do you want to enforce dailiness here too? i.e. once it has verified that the data has a date before or equal to the calibration start, should it call `assert_daily_data` on the calibration period?

dylanhmorris, line 404:
```suggestion
assert_sufficient_days_of_calibration_data <- function(date_vector,
```

R/checkers.R
```diff
@@ -317,3 +326,173 @@ assert_single_value <- function(x, arg = "x",
...
+      class = "wwinference_specification_error"
+    )
+  }
+  invisible()
+}
+
+#' Assert that a vector of dates we're testing overlaps with the comparison
+#' dates
+#'
+#' @param test_dates the vector of dates to check, must be in IS08 convention
```
dylanhmorris, line 428:
```suggestion
#' @param test_dates the vector of dates to check, must be strings in ISO8601 YY-MM-DD format
```

R/checkers.R
```diff
@@ -317,3 +326,173 @@ assert_single_value <- function(x, arg = "x",
...
+#' @param comparison_dates the vector of dates to compare to, must be in ISO8
+#' convention
```
dylanhmorris, line 430:
```suggestion
#' @param comparison_dates the vector of dates to compare to, must be in ISO8601 YY-MM-DD format
```

dylanhmorris: I'm not sure this is true? You don't call `lubridate::ymd` on the dates under the hood. Should you?

R/checkers.R
```diff
@@ -317,3 +326,173 @@ assert_single_value <- function(x, arg = "x",
...
+      ),
+      call = call,
+      class = "wwinference_specification_error"
+    )
+  }
+  invisible()
+}
+
+#' Assert that a vector of dates we're testing overlaps with the comparison
+#' dates
```
dylanhmorris, line 426: Can you explain a bit more what this function is for? What are `test` versus `comparison` dates

R/checkers.R
```diff
@@ -317,3 +326,173 @@ assert_single_value <- function(x, arg = "x",
...
+      call = call,
+      class = "wwinference_input_data_error"
+    )
+  }
+
+  invisible()
+}
+#' Assert that two tibbles of date and time mapping align
+#'
+#' @param full_data a tibble containing the columns `date` (with IS08 dates)
```
dylanhmorris, line 460:
```suggestion
#' @param full_data a tibble containing the columns `date` (with ISO8601 dates)
```

R/checkers.R
```diff
@@ -317,3 +326,173 @@ assert_single_value <- function(x, arg = "x",
...
+#' and `t` (integers of time in days) that the other tibble will be joined to
+#' @param subsetted_data a tibble containing the columns `date` (with
+#' IS08 dates) and `t` (integers of time in days) that we indexed based on the
```
dylanhmorris, line 463:
```suggestion
#' ISO8601 dates) and `t` (integers of time in days) that we indexed based on the
```

R/checkers.R
```diff
@@ -317,3 +326,173 @@ assert_single_value <- function(x, arg = "x",
...
+#'  `full_data`
+#' @param call Calling environment to be passed to [cli::cli_abort()] for
+#' traceback.
+#' @param add_err_msg add_err_msg string containing an additional error message,
+#' default is the empty string (`""`)
+#'
+#' @return NULL invisible
+assert_equivalent_indexing <- function(full_data,
```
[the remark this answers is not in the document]
dylanhmorris: Strikes me that we can make it more general and just check that any two dataframes are equivalently indexed (without requiring the user up front to specify that one is a subset of the other). Should we or is there a reason not to do this?

dylanhmorris (commented):
A few questions about the various checkers before I move on to reviewing the rest, @kaitejohnson

R/checkers.R
```diff
@@ -63,31 +63,32 @@ assert_elements_non_neg <- function(x, arg = "x",
...
-  if (checkmate::anyMissing(x)) {
-    cli::cli_abort(
-      c("{.arg {arg}} has missing values", add_err_msg,
-        "!" = "All elements of{.arg {arg}} should be present",
-        "i" = "Element(s) {.val {which(is.na(x))}} are missing"
-      ),
-      class = "wwinference_input_data_error",
-      call = call
-    )
+assert_integerish <-
```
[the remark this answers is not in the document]
kaitejohnson: Whoops yes!

R/checkers.R
```diff
@@ -317,3 +326,173 @@ assert_single_value <- function(x, arg = "x",
...
+      ),
+      call = call,
+      class = "wwinference_input_data_error"
+    )
+  }
+  invisible()
+}
+
+
+#' Assert that the vector of dates being passed in contains dates for each day
```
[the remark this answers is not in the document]
kaitejohnson: I don't think so -- that is a separate assertion (`assert_no_repreated_elements()`) which we will want to be able to toggle off if we at some point add a feature that allows the user to specify multiple admissions data streams. In this case, I want to ensure that there is at least one entry per day. Will update documentation accordingly.

R/checkers.R
```diff
@@ -317,3 +326,173 @@ assert_single_value <- function(x, arg = "x",
...
+#' @param call Calling environment to be passed to [cli::cli_abort()] for
+#' traceback.
+#' @param add_err_msg add_err_msg string containing an additional error message,
+#' default is the empty string (`""`)
+#'
+#' @return NULL invisible
+assert_daily_data <- function(dates,
+                              call = rlang::caller_env(),
+                              add_err_msg = "") {
+  sorted_dates <- dates[order(dates)]
```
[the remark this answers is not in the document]
kaitejohnson: good point

R/checkers.R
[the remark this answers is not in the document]
kaitejohnson: Nope good point!

R/checkers.R
kaitejohnson: Ah yeah, I think I can make this more general-- it just is essentially asserting that two sets of date vectors have some dates that overlap, and that both are before the max date.

R/checkers.R
kaitejohnson: I can add this.

R/checkers.R
kaitejohnson: I don't think so no, I want to keep these checks as separate as possible because for example, if we add a feature for weekly data, we can still have the user specify a calibration time in days, and then this will just check that theres enough data for that calibration time. A separate check will check for the daily nature of the data, and we can remove it for that use case. 

I'd rather keep these checks/asserts as modular as possible so that when we do make changes to the input data we know which ones we expect to fail.

kaitejohnson: too long, I think its ok to leave as is

R/checkers.R
kaitejohnson: Or actually, to keep this general I don't think we need that, we just want to check that they are dates. Added an `assert_date()`

kaitejohnson:
@dylanhmorris Think I addressed all of your questions here, ready for review

R/initialization.R
```diff
@@ -4,82 +4,84 @@
 #' @param stan_data a list of data elements that will be passed to stan
 #' @param params a dataframe of parameter values that are passed to stan
 #' to specify the priors in the model
+#' @param stdev a numeric value indicating the standard deviation to sample
+#' from when initializing, particularly from a standard normal. Also acts as
+#' a multiplier on the prior standard deviation, to restrict the initial value
+#' to be sampled from the center of the prior. Default is 0.01
```
dylanhmorris, line 10: I think this documentation needs rewording for clarity. If the goal of this function is to initialize each parameter near its prior mode, should say that explicitly, and also clarify that the distributional shapes specified below are the prior distribution shapes from the actual model. But it's not clear to me that that's exactly true in all cases.

Also, if the goal is to be less diffuse than the priors, why not provide a multiplicative factor for transforming the prior sds, rather than hard coding, i.e. something like

```R
my_scalar_param <- rnorm(1, params$my_param_prior_mean, scaling_factor * params$my_param_prior_sd)
```

R/initialization.R
```diff
@@ -4,82 +4,84 @@
...
 #' @return a list of initial values for each of the parameters in the
 #' `wwinference` model
-get_inits <- function(stan_data, params) {
-  # Assign parmeter names
-  par_names <- colnames(params)
-  for (i in seq_along(par_names)) {
-    assign(par_names[i], as.double(params[i]))
-  }
-
+get_inits <- function(stan_data, params, stdev = 0.01) {
```
dylanhmorris, line 14: As written this will only generate one chain worth of inits. Is the intended use something like  

```R
lapply(1:n_chains, function(x){get_inits(stan_data, params)})
```
?

If we're going to have a user-exported function why not have that functionality under the hood, take the number of chains as an argument, and then call a `get_inits_for_one_chain` helper function under the hood.

Doesn't have to be for this PR.

R/preprocessing.R
```diff
@@ -292,8 +293,10 @@ indicate_ww_exclusions <- function(data,
     # Port over the outliers flagged to the exclude column
     data_w_exclusions <- data |>
       dplyr::mutate(
-        exclude = ifelse({{ outlier_col_name }} == 1, 1, exclude)
+        exclude = ifelse(!!rlang::sym(outlier_col_name) == 1, 1, exclude)
```
dylanhmorris, line 296:
```suggestion
        exclude = ifelse(.data[[outlier_col_name]] == 1, 1, exclude)
```

R/wwinference.R
```diff
@@ -62,10 +62,7 @@
...
 wwinference <- function(ww_data,
                         count_data,
-                        model_spec = wwinference::get_model_spec(
-                          forecast_date =
-                            "2023-12-06"
-                        ),
+                        model_spec = wwinference::get_model_spec(),
                         mcmc_options = wwinference::get_mcmc_options(),
                         generate_initial_values = TRUE,
                         compiled_model = wwinference::compile_model()) {
```
dylanhmorris, line 68: Since these functions used for defaults are defined within the same package as the `wwinference()` function, they shouldn't need the `wwinference` prefix.

R/wwinference.R
```diff
@@ -62,10 +62,7 @@
 #'
 wwinference <- function(ww_data,
                         count_data,
-                        model_spec = wwinference::get_model_spec(
-                          forecast_date =
-                            "2023-12-06"
-                        ),
+                        model_spec = wwinference::get_model_spec(),
```
dylanhmorris, line 65: Perhaps I am missing something, but I do not think `model_spec` should have a default value. I don't see why `wwinference` should expect a particular `model_spec` by default any more than it expects particular `ww_data` or `count_data` values. 

Not for this PR, but I would in the long run suggest doing away with the `model_spec` construct. For example, I don't think the generation interval (as a vector) and the path to a parameter file (as a strong) should be passed to `wwinference` as part of the same list, while the `ww_data` and `count_data` are passed separately.

R/wwinference.R
dylanhmorris, line 68:
```suggestion
                        compiled_model = compile_model()) {
```

R/wwinference.R
```diff
@@ -62,10 +62,7 @@
...
                         mcmc_options = wwinference::get_mcmc_options(),
```
dylanhmorris, line 66:
```suggestion
                        mcmc_options = get_mcmc_options(),
```

R/wwinference.R
dylanhmorris, line 65:
```suggestion
                        model_spec,
```

dylanhmorris: Note that documentation will need to change too.

R/wwinference.R
```diff
@@ -87,22 +84,25 @@ wwinference <- function(ww_data,
     include_ww = as.numeric(model_spec$include_ww),
     compute_likelihood = 1
   )
+  # We want to return these to the user
+  input_data <- input_data_and_args$input_data
+  stan_args <- input_data_and_args$stan_args
```
dylanhmorris, line 89: Did tests pass here? I am concerned that as written it will no longer pass the ww and count data (which iirc are under `input_data`, not `stan_args`, to the `sample()` call. Am I missing something? If I am not and this is a potential bug, perhaps worth writing a test that would've caught this.

dylanhmorris (changes requested):
Some key questions before I review the tests themselves

R/get_stan_data.R
```diff
@@ -52,6 +121,15 @@ get_stan_data <- function(input_count_data,
     message("No wastewater data present")
   }
 
+  if (all(sum(input_ww_data$flag_as_ww_outlier) > sum(input_ww_data$exclude))) {
+    cli::cli_warn(
+      c(
+        "Wastewater data being passed to the model has outliers flagged,",
+        "but not all have been indicated for exclusion from model fit"
```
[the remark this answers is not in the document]
kaitejohnson: I had thought we had discussed this as something we wanted-- to separate the flagging of outliers and then the manual flagging of them for exclusion, as a separate step. This is how it is set up in the vignette, and was based on a conversation at a meeting a while ago, with the idea being that someone could visualize the data with flagged outliers and then choose to exclude those + others if they wanted to by adding to the exclusions column/running `indicate_ww_data_for_exclusion()`

R/get_stan_data.R
```diff
@@ -213,45 +334,54 @@ get_stan_data <- function(input_count_data,
...
   )
 
-  return(data_renewal)
+  input_data <- list(
+    input_ww_data = ww_data,
+    input_count_data = count_data
+  )
+  input_data_and_args <- list(
+    input_data = input_data,
+    stan_args = data_renewal
```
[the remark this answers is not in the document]
kaitejohnson: In this function, I generate the table of count data and wastewater data that is immediately passed to stan. My thought was that this is the data we want to return to the user in the `wwinference()` wrapper model function as is described in #49, and this is just the place where it is generated. Otherwise, there is no reason to have this nested list.

R/initialization.R
kaitejohnson: I think we could and maybe should do something like that, though of course some of them are not parameterized by their mean and sd or logmean and logsd (which is fin, just requires a bit more work in those instances)

I can try to rework this to make it more general, knowing that it will change the initializations slightly but that should be fine.

R/initialization.R
kaitejohnson: Yup, that is what happens I do a for loop in the `wwinference()` function. Can rename.

R/wwinference.R
kaitejohnson: Hmmm I am not sure I am completely following -- we had talked a while back about setting this up this way, similar to how `epinowcast`/`epinow2()` takes in a function that generates a list with defaults, so that instead of writing all the separate inputs (which we used to do in the other fitting functions in the `cfaforecastrenewalww` package, this just calls a function with defaults to generate those). My understanding was that this was better practice, especially since we are setting up this package to by default fit covid data and genome concentration data and all of the parameters/delay distributions being specified are specific to that.

https://github.com/epinowcast/epinowcast/blob/5aa7bcfae0766e14938ff1ad2c9c2f649ed7e86c/R/epinowcast.R#L106

Maybe I misunderstood that conversation but this was my attempt at implementing that. Tagging @seabbs and @gvegayon bc I think they were part of that convo

R/wwinference.R
kaitejohnson: Yes they passed, this is redundant bc `stan_args` contains the full set of data needed to be passed to stan. The `input_data` is just a nested list of the data that was passed to stan with dates bound to it. I wanted to make it a list so that as we increase the input data (e.g. and an adjacency matrix) this becomes easier to modify because the set of outputs from `wwinference()` will be stable. See #49 for this, sorry there is a bit of scope creep in this PR since I wanted to write the tests of `get_stan_data()` with the correct outputs

R/get_stan_data.R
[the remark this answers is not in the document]
dylanhmorris: Clarification: I think we should discuss whether automated detection/flagging of suspected outliers is in scope for this package at all (versus something we should let the user do for themself if they see fit).

R/get_stan_data.R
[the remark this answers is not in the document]
dylanhmorris: I see, thanks. If that's what you want, I would suggest splitting the activity in this `get_stan_data()` function:
- one function that creates the table currently returned as `input_data_and_args$input_data$input_ww_data`
- a second function that creates the table currently returned as `input_data_and_args$input_data$input_count_data
- A third function that takes those two things plus `generation_interval`, `inf_to_count_delay`, `infection_feedback_pmf`, and `params` as arguments, and creates a properly formatted data list to pass to `sample()`
I would then call those three functions in that order within `wwinference()`

dylanhmorris (commented):
A couple quick comments.

R/wwinference.R
gvegayon: I don't recall the conversation, but I am not against having defaults like this. This makes things easier when starting with the package and makes it even more appealing to new users. Nonetheless, perhaps it would be a good idea to (in a separate PR) add good **`print` and `summary` methods that explicitly clarify the model specs**.

An argument in favor of having defaults like this is that you have spent no trivial amount of time thinking about these numbers, so having a well-thought starting point is more beneficial than anything else to users.

kaitejohnson: Am going to keep this with default values for now!

R/get_stan_data.R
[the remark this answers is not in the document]
kaitejohnson: Ok I did my best to refactor this, I think one of the issues is that the wastewater data is sort of conditioned on the hospital admissions data (so had to change some things so that we're passing around the first and last date and stuff). Introduced a bit of redundancy that I am not sure is necessary/ideal.

R/wwinference.R
kaitejohnson: Yeah I think that is a good idea for #59

R/checkers.R
```diff
@@ -317,3 +323,186 @@ assert_single_value <- function(x, arg = "x",
...
+#' @return NULL invisible
+assert_overlap_dates <- function(dates1,
+                                 dates2,
+                                 max_date,
+                                 call = rlang::caller_env(),
+                                 add_err_msg = "") {
+  checkmate::assert_date(dates1)
+  checkmate::assert_date(dates2)
+  check_data_overlap <- min(dates1) <= max(dates2) &
+    max(dates1) <= max_date
```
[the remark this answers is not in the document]
kaitejohnson: Yeah, this is a great flag, this isn't named correctly. What I am really checking is that the second set of dates is within the first date in the first set of dates and the maximum date. Because for example we could have:
```r
hosp_dates<- lubridate::ymd(c("2023-01-01", "2023-01-02"))
ww_dates <- lubridate::ymd(c("2023-01-03", "2023-01-04"))
max_date <- "2023-01-05"
checkmate::assert_date(hosp_dates) # passes
checkmate::assert_date(ww_dates) # passes 
check_data_overlap <- min(hosp_dates) <= max(ww_dates) &
min(ww_dates) <= max(hosp_data) &
    max(ww_dates) <= max_date  &
max(hosp_dates) <=- max_date # TRUE
```

Will revise

kaitejohnson:
Ok I am a bit stuck on this particular CI issue, seems to have randomly cropped up. Otherwise ready for your continued review @dylanhmorris

R/initialization.R
```diff
@@ -4,82 +4,84 @@
...
     sigma_ww_site_sd = abs(stats::rnorm(
-      1, sigma_ww_site_prior_sd_mean,
-      0.1 * sigma_ww_site_prior_sd_sd
+      1, params$sigma_ww_site_prior_sd_mean,
+      stdev * params$sigma_ww_site_prior_sd_sd
     )),
-    sigma_ww_site_raw = abs(stats::rnorm(n_ww_lab_sites, 0, 0.05)),
-    p_hosp_mean = stats::rnorm(1, stats::qlogis(p_hosp_mean), 0.01),
-    p_hosp_w = stats::rnorm(tot_weeks, 0, 0.01),
+    sigma_ww_site_raw = abs(stats::rnorm(n_ww_lab_sites, 0, stdev)),
```
gvegayon, line 68: This replacement will change the outcomes of the code; was this your intent?

R/initialization.R
```diff
@@ -4,82 +4,84 @@
...
+      1, params$duration_shedding_mean,
+      stdev * params$duration_shedding_sd
     ),
-    log10_g = stats::rnorm(1, log10_g_prior_mean, 0.5),
-    ww_site_mod_raw = abs(stats::rnorm(n_ww_lab_sites, 0, 0.05)),
-    ww_site_mod_sd = abs(stats::rnorm(1, 0, 0.05)),
-    hosp_wday_effect = to_simplex(stats::rnorm(7, 1 / 7, 0.01)),
+    log10_g = stats::rnorm(1, params$log10_g_prior_mean, 0.5),
+    ww_site_mod_raw = abs(stats::rnorm(n_ww_lab_sites, 0, stdev)),
+    ww_site_mod_sd = abs(stats::rnorm(1, 0, stdev)),
```
gvegayon, line 83: Smake on these two. It used to be 0.05, now it is 0.01.

R/utils.R
```diff
@@ -144,3 +144,18 @@ convert_to_logsd <- function(mean, sd) {
...
+#' @return A string vector where all single brackets are replaced with double
+#' brackets
+#' @export
+#'
+#' @examples
+#' test_string <- "{my_var}"
+#' glue::glue(autoescape_brackets(test_string))
+autoescape_brackets <- function(string) {
+  return(gsub("\\{(.*)\\}", "{{\\1}}", string))
+}
```
gvegayon, line 161: Shouldn't this be an internal function?

R/wwinference.R
```diff
@@ -254,7 +268,7 @@ get_mcmc_options <- function(
 #'
 #'
 #' @param forecast_date a character string in ISO8601 format (YYYY-MM-DD)
-#' indicating the date that the forecast is to be made. Default is
+#' indicating the date that the forecast is to be made. Default is NULL
```
gvegayon, line 271: What happens when the default is `NULL`? Will there be no forecast?

gvegayon (commented):
Looks great, @kaitejohnson! A few minor comments. Since you are adding testing, I wonder: Do you have the coverage information? That should be handy now.

R/utils.R
dylanhmorris: Agree, I would not export this.

R/initialization.R
dylanhmorris: Yes, @kaitejohnson intended to change this. Longer term I would like to refactor the init simulation a bit but that's a separate PR.

R/initialization.R
dylanhmorris: See https://github.com/CDCgov/ww-inference-model/pull/54/files#r1717397110

R/utils.R
kaitejohnson: I might have needed it for an example of something I was exporting (which probably also could be an external function-- let me look!)

tests/testthat/test_checkers.R
```diff
@@ -0,0 +1,214 @@
...
+    )
+    expect_no_error(assert_daily_data(daily_dates))
+
+    weekly_dates <- c(
+      lubridate::ymd("2023-01-01"),
+      lubridate::ymd("2023-01-08")
+    )
+    expect_error(assert_daily_data(weekly_dates))
+  }
+)
```
dylanhmorris, line 214: Test for the dates in range function?

R/wwinference.R
kaitejohnson: Yeah so my goal here was to follow how some other R packages have a default of NULL when something is required. If the forecast date is missing, then the function errors with an informative error message telling the user to specify a forecast date. Almost everything else can follow defaults, but I think this should be required. 

See here: https://github.com/CDCgov/ww-inference-model/blob/48ac3a1c427154c6bd31f9dbabb3aed53ce6a849/R/wwinference.R#L312

R/initialization.R
kaitejohnson: Yep I tried to do some scope creep and make this a bit more generic per #20 . Need to do more of a refactor where we sample from some constrained space centered around the center of the prior across all parameters, but for now this was my attempt at not hard coding all of the inits...

tests/testthat/test_get_stan_data.R
```diff
@@ -0,0 +1,214 @@
...
+    calibration_time,
+    generation_interval,
+    inf_to_count_delay,
+    infection_feedback_pmf,
+    params,
+    include_ww
+  ))
+})
+
+test_that("Test that pmf check works as expected", {
```
dylanhmorris, line 162: Would be good to have a unit test for the PMF validation function itself in`test_checkers.R`. But checking that it gets called here and behaves as expected in context is nice to have too.

tests/testthat/test_indicate_ww_exclusions.R
```diff
@@ -0,0 +1,42 @@
...
+test_that("No rows are marked for exclusion when remove_outliers is FALSE", {
+  processed <- indicate_ww_exclusions(data,
+    outlier_col_name = "flag_as_ww_outlier",
+    remove_outliers = FALSE
+  )
+
+  expected_exclude <- c(0, 0) # No rows should be excluded
+
+  expect_equal(processed$exclude, expected_exclude)
+})
```
dylanhmorris, line 42: We should also test that indicated excludes are indeed removed by the removal function and that things not marked for exclusion are _not_ removed. Is there a test for that anywhere?

tests/testthat/test_indicate_ww_exclusions.R
```diff
@@ -0,0 +1,42 @@
...
+)
+
+# Test that function returns a dataframe with an additional 'exclude' column
+test_that("Function returns dataframe with 'exclude' column", {
+  processed <- indicate_ww_exclusions(data,
+    outlier_col_name = "flag_as_ww_outlier",
+    remove_outliers = TRUE
+  )
+
+  expect_true("exclude" %in% names(processed))
```
dylanhmorris, line 16: Use `checkmate::expect_names`?

tests/testthat/test_preprocess_count_data.R
```diff
@@ -0,0 +1,176 @@
...
+# Test that function returns a dataframe with correct columns
+test_that("Function returns dataframe with correct columns", {
+  processed <- preprocess_count_data(hosp_data,
+    count_col_name = "daily_admits",
+    pop_size_col_name = "state_pop"
+  )
+
+  expected_cols <- c("date", "count", "total_pop")
+
+  expect_true(all(expected_cols %in% names(processed)))
```
dylanhmorris, line 17: Use `checkmate::expect_names`?

tests/testthat/test_preprocess_count_data.R
```diff
@@ -0,0 +1,176 @@
...
+
+# Test that count column is renamed correctly
+test_that("Count column is renamed correctly", {
+  processed <- preprocess_count_data(hosp_data,
+    count_col_name = "daily_admits",
+    pop_size_col_name = "state_pop"
+  )
+
+  expect_false("daily_admits" %in% names(processed))
+  expect_true("count" %in% names(processed))
```
dylanhmorris, line 28: Use `checkmate::expect_names`?

tests/testthat/test_preprocess_count_data.R
```diff
@@ -0,0 +1,176 @@
...
+
+# Test that population size column is renamed correctly
+test_that("Population size column is renamed correctly", {
+  processed <- preprocess_count_data(hosp_data,
+    count_col_name = "daily_admits",
+    pop_size_col_name = "state_pop"
+  )
+
+  expect_false("state_pop" %in% names(processed))
+  expect_true("total_pop" %in% names(processed))
```
dylanhmorris, line 39: Use `checkmate::expect_names`?

tests/testthat/test_preprocess_count_data.R
```diff
@@ -0,0 +1,176 @@
...
+# Test that function returns a dataframe with correct columns
+test_that("Function returns dataframe with correct columns", {
+  processed <- preprocess_count_data(hosp_data,
+    count_col_name = "daily_admits",
+    pop_size_col_name = "state_pop"
+  )
+
+  expected_cols <- c("date", "count", "total_pop")
+
+  expect_true(all(expected_cols %in% names(processed)))
```
dylanhmorris, line 58: Use `checkmate::expect_names`?

tests/testthat/test_preprocess_count_data.R
```diff
@@ -0,0 +1,176 @@
...
+
+# Test that count column is renamed correctly
+test_that("Count column is renamed correctly", {
+  processed <- preprocess_count_data(hosp_data,
+    count_col_name = "daily_admits",
+    pop_size_col_name = "state_pop"
+  )
+
+  expect_false("daily_admits" %in% names(processed))
+  expect_true("count" %in% names(processed))
```
dylanhmorris, line 69: Use `checkmate::expect_names`?

tests/testthat/test_preprocess_count_data.R
```diff
@@ -0,0 +1,176 @@
...
+
+# Test that population size column is renamed correctly
+test_that("Population size column is renamed correctly", {
+  processed <- preprocess_count_data(hosp_data,
+    count_col_name = "daily_admits",
+    pop_size_col_name = "state_pop"
+  )
+
+  expect_false("state_pop" %in% names(processed))
+  expect_true("total_pop" %in% names(processed))
```
dylanhmorris, line 80: Use `checkmate::expect_names`?

kaitejohnson:
> Looks great, @kaitejohnson! A few minor comments. Since you are adding testing, I wonder: Do you have the coverage information? That should be handy now.

Yeah, let me look into adding that into CI?

tests/testthat/test_preprocess_count_data.R
```diff
@@ -0,0 +1,176 @@
...
+})
+
+test_that("Function handles missing columns with an error", {
+  incomplete_hosp_data <- hosp_data[, -which(names(
+    hosp_data
+  ) == "daily_admits")]
```
dylanhmorris, line 86:
```suggestion
  incomplete_hosp_data <- hosp_data |> dplyr::select(-daily_admits)
  
```

tests/testthat/test_preprocess_ww_data.R
```diff
@@ -0,0 +1,279 @@
...
+# Test that no NA values are introduced during preprocessing
+# (assuming input has no NAs)
+test_that("No NA values are introduced during preprocessing", {
+  processed <- preprocess_ww_data(ww_data,
+    conc_col_name = "conc",
+    lod_col_name = "lod"
+  )
+
+  # Check for any new NA values introduced in any column
+  expect_true(all(complete.cases(processed)))
```
dylanhmorris, line 263: What the reason for `complete.cases` over `!is.na`? Open to it but would like to know.

dylanhmorris (changes requested):
Coming along well, thanks @kaitejohnson. A few things.

R/validate.R
```diff
@@ -130,3 +144,96 @@ validate_count_data <- function(hosp_data,
...
+  checkmate::assert_integerish(calibration_time)
+
+  # make sure filtering to exclude days before earliest calibration time
+  # didn't eliminate
+  assert_df_not_empty(input_ww_data,
+    add_err_msg = c(
+      "Wastewater data passed in doesn't overlap",
+      "with count data calibration period"
+    )
+  )
```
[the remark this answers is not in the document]
kaitejohnson: I flipped these two error messages

R/validate.R
```diff
@@ -130,3 +144,96 @@ validate_count_data <- function(hosp_data,
...
+  # data
+  assert_dates_within_frame(
+    input_count_data$date,
+    input_ww_data$date,
+    forecast_date,
+    add_err_msg = c(
+      "There is no wastewater data within ",
+      "the count data calibration period"
+    )
+  )
```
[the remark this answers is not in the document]
kaitejohnson: The error message is wrong, see my comment above. Just flipped. 

Yes, this is not a general function I wouldn't say. it's to check that one set of dates fits within the min of another and a max date, which is really only relevant for checking that ww data is within the frame in this context.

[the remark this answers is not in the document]
kaitejohnson: > Should this always be an error? Or should there be a require_wastewater flag or something? What if a user who doesn't have any wastewater data wants to get posterior predictive wastewater dynamics in light of their count_data?

Currently the package requires wastewater data. You can fit the model without it by turning it off, but I have to do some work to not fit to the wastewater data bc its not present. I think this should be in a separate PR.  As part of that workflow, none of the calls to validate_ww_data or validate_both_datasets would be invoked

R/wwinference.R
```diff
@@ -92,17 +106,17 @@ wwinference <- function(ww_data,
...
     for (i in 1:mcmc_options$n_chains) {
-      init_lists[[i]] <- get_inits(stan_data, params)
+      init_lists[[i]] <- get_inits_for_one_chain(stan_args, params)
     }
   }
 
 
   fit_model <- function(compiled_model,
-                        standata,
+                        stan_args,
```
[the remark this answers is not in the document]
kaitejohnson: I made the choise to switch to `stan_args` in part because I want to return to the user both the "data" passed to stan (these are dataframes in the `input_data` list in a subsequent PR #59 ), and the `stan_args` which is a nested list with only vectors, matrices, scalars, integers etc. I think the reason I prefer args is to distinguish between the many things we are calling `data` which are almost always in the form of a dataframe/tibble.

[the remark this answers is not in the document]
kaitejohnson: I think args is general enough as arguments. I think of configuration arguments as specifications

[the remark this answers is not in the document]
dylanhmorris: Counterpoint: this list is specifying the thing called the `data` block in a Stan model and is being passed as the `data` named argument to `sample()`. Calling it something with `data` in it makes that clearer. Can we compromise on `stan_data_list` as suggested below? 

If you want to distinguish it from processed but not `sample()`-`data`-argument-formatted data tables, I think calling the others `input_ww_data` or `input_count_data` versus a `stan_data_list` suffices.

tests/testthat/test_indicate_ww_exclusions.R
kaitejohnson: No I can add that

tests/testthat/test_preprocess_ww_data.R
kaitejohnson: No reason, I think this is one chat gpt suggested for me :)

kaitejohnson:
Ok @dylanhmorris I think I covered all of your suggestions/additional tests. I think any other tests should be new issues, this was mostly just to have some amount of testing infrastructure, not to be fully comprehensive and I am sure we will find bugs/more things to write tests for as we go. 

@gvegayon I am going to look into including testing coverage in CI

R/utils.R
dylanhmorris: I'm guessing it's the example you wrote here for the function itself. The catch is that internal functions typically [don't have examples](https://github.com/tidyverse/dplyr/blob/main/R/utils.R).

```R
> test_string <- "{my_var}"
> glue::glue(autoescape_brackets(test_string))
Error in autoescape_brackets(test_string) : 
  could not find function "autoescape_brackets"
```

tests/testthat/test_preprocess_count_data.R
```diff
@@ -0,0 +1,185 @@
...
+# Test that no NA values are introduced during preprocessing
+# (assuming input has no NAs)
+test_that("No NA values are introduced during preprocessing", {
+  processed <- preprocess_count_data(hosp_data,
+    count_col_name = "daily_admits",
+    pop_size_col_name = "state_pop"
+  )
+
+  # Check for any new NA values introduced in any column
+  expect_true(all(complete.cases(processed)))
```
dylanhmorris, line 122:
```suggestion
  expect_false(any(is.na(processed)))
```

tests/testthat/test_preprocess_count_data.R
```diff
@@ -0,0 +1,185 @@
...
+    daily_hosp_admits = c(10, 20),
+    state_pop = c(1e6, 1e6)
+  )
+  # Using default column names
+  default_processed <- preprocess_count_data(test_hosp_data)
+
+  expect_false(any(c(
+    "daily_admits", "state_pop"
+  ) %in% names(default_processed)))
+  expect_true(all(c("count", "total_pop") %in% names(default_processed)))
```
dylanhmorris, line 138:
```suggestion
  checkmate::expect_names(
      names(default_processed), 
      must.include = c("count", "total_pop") , 
      disjoint.from = c("daily_admits", "state_pop"))
```

dylanhmorris (changes requested):
Very close. A few remaining typos and places where things can be simplified.

R/validate.R
```diff
@@ -202,7 +201,9 @@ validate_both_datasets <- function(input_count_data,
   invisible()
 }
 
-#' Validate that the pmf being passed to stan
+#' Validate that the pmf vector being passed to stan
+#' is a valid probability mass function sums to 1 and
+#' has all non-negative entries.
```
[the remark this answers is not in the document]
kaitejohnson: pmf vector is singular, so I think this should be has since that is the object its referring to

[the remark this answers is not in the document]
dylanhmorris: This is in conjuction with splitting into to sentences above. Goal is to end up with "It must sum to 1 and have all non-negative entries."

tests/testthat/test_preprocess_count_data.R
kaitejohnson: OK just confirmed that this does in fact error. The argument is called `disjunct.from = ` not `disjoint.from `

tests/testthat/test_preprocess_ww_data.R
```diff
@@ -0,0 +1,280 @@
...
+
+# Test that concentration column is renamed correctly
+test_that("Concentration column is renamed correctly", {
+  processed <- preprocess_ww_data(ww_data,
+    conc_col_name = "conc",
+    lod_col_name = "lod"
+  )
+
+  expect_false("conc" %in% names(processed))
+  expect_true("genome_copies_per_ml" %in% names(processed))
```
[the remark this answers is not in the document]
kaitejohnson: other way around

R/wwinference.R
```diff
@@ -62,21 +62,35 @@
 #'
 wwinference <- function(ww_data,
                         count_data,
-                        model_spec = wwinference::get_model_spec(
-                          forecast_date =
-                            "2023-12-06"
-                        ),
-                        mcmc_options = wwinference::get_mcmc_options(),
+                        model_spec = get_model_spec(),
```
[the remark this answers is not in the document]
kaitejohnson: Ok this was a point I have a question on: should I just separate forecast date from the rest of the model specifications? It is the one thing that I really want the model to fail loudly on if a user doesn't provide a forecast date. So as is, if the user doesn't put a forecast date into `get_model_spec()` it will error with an informative error message. Should I just move forecast date out of the `get_model_spec()` function? I wasn't sure how to handle this

R/wwinference.R
```diff
@@ -294,6 +308,11 @@ get_model_spec <- function(
         package = "wwinference"
       )
     )) {
+  if (is.null(forecast_date)) {
```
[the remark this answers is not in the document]
kaitejohnson: I initially had had a default forecast date but that seemed like a really bad idea. See my comment above. It seemed unnecessary to move to a separate argument, but I do think it is the one specification that I want to require the user to provide

tests/testthat/test_preprocess_count_data.R
dylanhmorris: Apologies, typing from memory!

R/wwinference.R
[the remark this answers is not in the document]
dylanhmorris: > Should I just move forecast date out of the get_model_spec() function? I wasn't sure how to handle this

This is what I would do in the first instance.

tests/testthat/test_preprocess_count_data.R
```diff
@@ -0,0 +1,184 @@
...
+
+# Test that count column is renamed correctly
+test_that("Count column is renamed correctly", {
+  processed <- preprocess_count_data(hosp_data,
+    count_col_name = "daily_admits",
+    pop_size_col_name = "state_pop"
+  )
+
+  expect_false("daily_admits" %in% names(processed))
+  expect_true(checkmate::check_names(names(processed), must.include = "count"))
```
dylanhmorris, line 76:
```suggestion
checkmate::expect_names(names(processed), must.include = "count", disjunct.from = "daily_admits")
```

tests/testthat/test_preprocess_count_data.R
```diff
@@ -0,0 +1,184 @@
...
+test_that("Function returns dataframe with correct columns", {
+  processed <- preprocess_count_data(hosp_data,
+    count_col_name = "daily_admits",
+    pop_size_col_name = "state_pop"
+  )
+
+  expected_cols <- c("date", "count", "total_pop")
+  expect_true(checkmate::check_names(names(processed),
+    must.include = expected_cols
+  ))
```
dylanhmorris, line 65:
```suggestion
  checkmate::expect_names(names(processed),
    must.include = expected_cols)
```

dylanhmorris (changes requested):
Re-flagging a couple `checkmate::expect_names` changes from the last review. Otherwise good to go @kaitejohnson

tests/testthat/test_preprocess_count_data.R
kaitejohnson: Oh my bad I initially wrapped these bc I thought they worked like `check` and just returned TRUE or a string error message. But this function luckily does error. I will note that CI views this as an empty test because it doesn't use test that infrastructure I think

dylanhmorris: Interesting. `checkmate`'s `expect_` functions are designed to work with `testthat`. I wonder if this would still be viewed as an empty test if you ran `library(checkmate)` after `library(testthat)` at the top of your test.

Another workaround would just be to add in a `expect_true(TRUE)`, but that's a bit of an unpleasant workaround.

https://mllg.github.io/checkmate/articles/checkmate.html#checkmate-as-testthat-extension

dylanhmorris (dismissed):
LGTM. Feel free to merge. Did you want to move the `forecast_date` out of the `model_spec` in this PR or keep it for a separate one, @kaitejohnson.

kaitejohnson:
> LGTM. Feel free to merge. Did you want to move the `forecast_date` out of the `model_spec` in this PR or keep it for a separate one, @kaitejohnson.

I think it would be good to do in this PR! Sorry I must have missed a comment, do you think it makes sense to just make it a separate argument?

tests/testthat/test_preprocess_count_data.R
kaitejohnson: Ok now I am not seeing it anymore so maybe it was another issue

dylanhmorris:
> I think it would be good to do in this PR! Sorry I must have missed a comment, do you think it makes sense to just make it a separate argument?

Yes I do.

dylanhmorris (approved):
LGTM provided CI passes.

PR #58, opened 2024-08-09
Adding class and methods for wwinference model fit

gvegayon:
This PR does the following:

- Add `get_draws_df()` (new generic with S3 method: default and `wwinference_fit`)
- Adds `get_model_diagnostic_flags()` (new generic with S3 methods: default and `wwinference_fit`)
- Reduces the output from `wwinferece()` and wraps it as an object of class `wwinference_fit`.
- Adds `print.wwinference_fit()` (S3 method): Prints some information about the model.
- Adds `summary.wwinference_fit()` (S3 method): Calls the S6 summary method from the CmdStanModel.


R/wwinference.R
```diff
@@ -189,7 +189,17 @@ wwinference <- function(ww_data,
...
   }
 
-  return(out)
+  structure(out, class = "wwinference_fit")
+}
+
+#' @export
+print.wwinference_fit <- function(x, ...) {
+  cat("wwinference_fit object\n")
+  cat("Model fit object: ", x$fit, "\n")
```
kaitejohnson, line 198: This doesn't work on list objects I don't think!

gvegayon:
OK, @kaitejohnson, this is ready for review. To avoid getting stuck, I suggest adding more S3 methods later (e.g., `plot`, `vcov`, `confint`, etc.).

R/wwinference.R
gvegayon: Addressed in [4c7105b](https://github.com/CDCgov/ww-inference-model/pull/58/commits/4c7105bbda61a84c2401eb13a3bef6c2b49b1f96).

R/wwinference.R
```diff
@@ -41,31 +41,102 @@
...
+#' `fit`: The CmdStan object that is returned from the call to
 #' `cmdstanr::sample()`. Can be used to access draws, summary, diagnostics, etc.
-#' `date_time_spine`: Mapping from time in stan to dates
-#' `lab_site_spine`: Mapping from lab_site_index in stan to lab and site
-#' `subpop_spine`: Mapping from site index in stan to site
+#' `input_data`: a list containing the input `ww_data` and the input
+#' `count_data` used in the model.
+#' `stan_data_args`: a list containing the inputs passed directly to the
+#' stan model
+#' `mcmc_options`: a list of the MCMC specifications passed to stan
```
kaitejohnson, line 55: @seabbs Curious your thoughts on these outputs of the `wwinference()` wrapper function, as these would be what I would hope would stay stable even as we continue to develop on this.

R/wwinference.R
```diff
@@ -218,9 +231,69 @@ wwinference <- function(ww_data,
...
+#' @param ... Additional parameters passed to the corresponding method
+#' @export
+#' @rdname wwinference
+#' @return
+#' - The print method prints out information about the model and
+#' returns the object invisibly.
+print.wwinference_fit <- function(x, ...) {
+  cat("wwinference_fit object\n")
+  cat("N of WW sites    :", x$stan_args$n_ww_sites, "\n")
+  cat("N of lab sites   :", x$stan_args$n_ww_lab_sites, "\n")
```
kaitejohnson, line 248:
```suggestion
  cat("N of unique lab-site pairs  :", x$stan_args$n_ww_lab_sites, "\n")
```

kaitejohnson (commented):
@gvegayon This looks good to me, my only suggestion would be that we add the summary to the `wwinference.Rmd`! Also would like some feedback from @seabbs and @dylanhmorris if the outputs of `wwinference()` look good to them, since I think we would want to try to keep this as consistent as possible going forward (so as not to introduce breaking changes with new releases).

seabbs:
> Add get_draws_df() (new generic with S3 method: default and wwinference_fit)

if going to do this why not write a method for `posterior`

R/model_diagnostics.R
```diff
@@ -18,24 +21,46 @@
...
+#' @export
+get_model_diagnostic_flags.wwinference_fit <- function(x, ...) {
+  get_model_diagnostic_flags.default(
+    x = x$fit$result
+  )
+}
+
+#' @rdname get_model_diagnostic_flags
+#' @export
+get_model_diagnostic_flags.default <- function(x,
```
seabbs, line 51: is this really a default method or is it a cmdstan method?

R/model_diagnostics.R
```diff
@@ -57,5 +82,37 @@ get_model_diagnostic_flags <- function(stan_fit_object,
...
+
+#' Method for printing the CmdStan parameter diagnostics for a
+#' wwinference_fit_object
+#'
+#' @param ww_fit An object of class wwinference_fit
+#' @param ... additional arguments
+#'
+#' @family diagnostics
+#' @export
+parameter_diagnostics <- function(ww_fit, ...) {
```
seabbs, line 116: why do these nearly the same functions exist?

R/wwinference.R
```diff
@@ -41,31 +41,102 @@
...
-#' observation to facilitate plotting.
-#' `raw_fit_obj`: The CmdStan object that is returned from the call to
+#' `fit`: The CmdStan object that is returned from the call to
 #' `cmdstanr::sample()`. Can be used to access draws, summary, diagnostics, etc.
-#' `date_time_spine`: Mapping from time in stan to dates
-#' `lab_site_spine`: Mapping from lab_site_index in stan to lab and site
-#' `subpop_spine`: Mapping from site index in stan to site
+#' `input_data`: a list containing the input `ww_data` and the input
+#' `count_data` used in the model.
+#' `stan_data_args`: a list containing the inputs passed directly to the
```
seabbs, line 53: it feels like the names here are inconsistent. 

I think mcmc_options should be fit_opts or similar as this allows you to generalise if you don't use mcmc. I usually go with data_llist for stan_data_args but personal preference

R/wwinference.R
```diff
@@ -218,9 +231,69 @@ wwinference <- function(ww_data,
     }
   }
 
-  return(out)
+  structure(out, class = "wwinference_fit")
```
seabbs, line 234: need a class validator or similar

R/model_diagnostics.R
kaitejohnson: This is a custom function that we set up when we made the pipeline. It uses diagnostic thresholds that we set as defaults (user can alter them), and returns TRUE if the threshold is exceeded.

R/get_draws_df.R
```diff
@@ -5,38 +5,83 @@
...
+    x = x$input_data$input_ww_data,
+    count_data = x$input_data$input_count_data,
+    stan_args = x$stan_args,
+    fit_obj = x$fit
+  )
+}
+
+#' @rdname get_draws_df
+#' @export
+get_draws_df.default <- function(x,
```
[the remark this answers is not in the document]
kaitejohnson: This is if someone wants to pass in the tibbles themselves rather than the `wwinference` class object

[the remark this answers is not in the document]
kaitejohnson: I guess we could get rid of this and parse the args from the `wwinference` object but I think I have a slight preference for the default being that we pass in the individual components of rather than a class object.

R/model_diagnostics.R
kaitejohnson: Good point @gvegayon this is redundant from the function you created (`diagnostic_summary()`). I labeled this parameter diagnostics bc it helps me remember that this is the call that returns the diagnostics at the level of the individual parameter as opposed to the call to `cmdstan_obj$summary` that returns summaries across all parameters. Think we should just have one

R/wwinference.R
kaitejohnson: Can change to `fit_opts` that makes sense. 
I have slight preference for `stan_data_args` bc we return an `input_data` list (`ww_data`, `hosp_data`, and eventually `dist_matrix` which might be confused with `data_list`).

R/wwinference.R
kaitejohnson: Example?

kaitejohnson:
@gvegayon Reminder to add to `NEWS.md` with a description of this change! I keep forgetting to do that/keep the date updated.

R/wwinference.R
seabbs: I think I might call the input data `raw_` but I don't feel strongly

R/wwinference.R
seabbs: https://adv-r.hadley.nz/s3.html

13.3

R/get_draws_df.R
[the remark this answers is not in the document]
seabbs: then why isn't that a .dataframe method?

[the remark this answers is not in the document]
seabbs: The way that this is currrently coded literally anything people try and input gets passed to .default. If you instead had .default be say empty you could avoid some fairly large amount of input checks (as the message can be in .default)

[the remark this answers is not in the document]
seabbs: all of this aside for this actual function it seems like expanding  on posterior / on of the other standard names from stan land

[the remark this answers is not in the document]
kaitejohnson: I think the difference between this and `posterior::as_draws_df()` is that this just returns a few of the variables (observables + R(t) estimates) bound to dates/data if available, using `tidybayes()`. Perhaps I should rename to something more descriptive re `extract_posterior_draws_w_data()` ? Open to changing things around as now is the time to nail this down! But I do think this table, or something similar, is a useful output since this is what someone would use for plotting, or computing quantiles, etc.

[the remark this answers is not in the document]
kaitejohnson: > then why isn't that a .dataframe method?
 Bc I didn't know this was a thing. 

To recap your suggestion: 
1. change default to be empty (so just `...`?)
2. add a `.dataframe` method that takes in the necessary dataframes, currently described in `.default`
3. add a message in `.default()` telling the user to either pass the required dataframe args or the `wwinference_fit` object?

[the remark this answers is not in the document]
gvegayon: Default method now fails (stop) with message indicating the available methods.

R/model_diagnostics.R
gvegayon: Sounds good, keeping parameter_diagnostics then.

R/wwinference.R
gvegayon: So `input_data` -> `fit_opts` and `stan_data_args` stays?

dylanhmorris: I'd propose `stan_data_list` instead of `stan_data_args`

R/wwinference.R
gvegayon: Adding a constructor. Right now, only checking for object types. Can do fancier checking in a next PR.

R/wwinference.R
kaitejohnson: Hmm no I think we want (and @gvegayon I will do this as a PR into this branch):
`mcmc_options` -> `fit_opts`
`stan_data_args` -> `stan_data_list`
`input_data` -> `raw_input-data`

NEWS.md
```diff
@@ -1,6 +1,8 @@
 # wwinference
 
 This will serve as our change-log
+
+- 2024-08-23: Added new `wwinference_fit` class with corresponding print and summary methods.
```
kaitejohnson, line 5: I've been putting them in chronological order top to bottom!

vignettes/wwinference.Rmd
```diff
@@ -297,33 +309,62 @@ fit <- wwinference::wwinference(
     infection_feedback_pmf = infection_feedback_pmf,
     params = params
   ),
-  mcmc_options = get_mcmc_options(),
+  fit_opts = get_mcmc_options(iter_warmup = 100, iter_sampling = 100),
```
kaitejohnson, line 312: This makes sense-- confirming though that it didn't result in diagnostic issues?

vignettes/wwinference.Rmd
```diff
@@ -297,33 +309,62 @@ fit <- wwinference::wwinference(
...
-```{r}
-head(fit)
+# The `wwinference_fit` object
+
+The `wwinference()` function returns a `wwinference_fit` object which includes
+the underlying and the underlying
+[`CmdStanModel` object](https://mc-stan.org/cmdstanr/reference/CmdStanModel.html)
+ (`fit`), a list of the two sources of input
+data (`input_data`), the list of the arguments passed to stan
+(`stan_data_args`), and the list of the MCMC options (`mcmc_options`) passed to
```
kaitejohnson, line 324:
```suggestion
(`stan_data_list`), and the list of the MCMC options (`fit_opts`) passed to
```

vignettes/wwinference.Rmd
```diff
@@ -297,33 +309,62 @@ fit <- wwinference::wwinference(
...
-including the site-level R(t) estimates and the state-level R(t) estimate.
-```{r}
-head(fit)
+# The `wwinference_fit` object
+
+The `wwinference()` function returns a `wwinference_fit` object which includes
+the underlying and the underlying
+[`CmdStanModel` object](https://mc-stan.org/cmdstanr/reference/CmdStanModel.html)
+ (`fit`), a list of the two sources of input
+data (`input_data`), the list of the arguments passed to stan
```
kaitejohnson, line 323:
```suggestion
data (`raw_input_data`), the list of the arguments passed to stan
```

vignettes/wwinference.Rmd
```diff
@@ -297,33 +309,62 @@ fit <- wwinference::wwinference(
...
-We can create the plots using the fitting wrapper functions (figures can
-also created directly using the `draws_df`)
+### Using explicit passed arguments rather than S3 methods
+
+Rather than using S3 methods supplied for `wwinference()`, the elements in the
+`wwinference_fit` object can also be used directly to create this dataframe.
+This is demonstrated below:
+
+```{r extracting-draws-explicit}
+draws_df_explicit <- get_draws_df(
```
kaitejohnson, line 352: @gvegayon Question for my understanding -- I thought that this would call the `.default` method, since we haven't specified the `.dataframe` method. Wouldn't that mean that this should throw an error?

R/get_draws_df.R
```diff
@@ -5,38 +5,95 @@
...
+    x = x$raw_input_data$input_ww_data,
+    count_data = x$raw_input_data$input_count_data,
+    stan_data_list = x$stan_data_list,
+    fit_obj = x$fit
+  )
+}
+
+#' @export
+#' @rdname get_draws_df
+get_draws_df.default <- function(x, ...) {
```
kaitejohnson, line 53: When would this result in an error?

vignettes/wwinference.Rmd
kaitejohnson: But I think we should add some language around this saying that we used a smaller number of warmup and sampling iterations here as an example then we would recommend for production?

vignettes/wwinference.Rmd
```diff
@@ -283,9 +296,8 @@ set the MCMC settings using `get_mcmc_options()`, and pass in our
 pre-compiled model(`model`) to `wwinference()` where they are combined and
 used to fit the model.
 
-
-```{r warning=FALSE, message=FALSE}
-fit <- wwinference::wwinference(
+```{r fitting-model, warning=FALSE, message=FALSE}
+ww_fit <- wwinference::wwinference(
```
kaitejohnson, line 300: Can we call the `print()` method and `summary()` method to demonstrate how they work?

kaitejohnson (changes requested):
@gvegayon A few suggested changes and some questions. Overall looking really good and excited to have this working! I think I am just trying to wrap my head around what the use case of the `.default` is

R/get_draws_df.R
```diff
@@ -5,38 +5,95 @@
...
+#' S3 method for extracting posterior draws alongside data for a
+#' wwinference_fit object
+#'
+#' This method overloads the generic get_draws_df function specifically
+#' for objects of type 'wwinference_fit'.
+#'
+#' @rdname get_draws_df
+#' @export
+get_draws_df.wwinference_fit <- function(x, ...) {
+  get_draws_df.data.frame(
```
seabbs, line 43: Nice

R/get_draws_df.R
seabbs: it should do so if called on anything that isn't a dataframe (or inherits from one) or a `wwinference_fit` object (and if we don;'t we should have a test to check this)

R/model_diagnostics.R
seabbs: no sorry this is the same question as for the other s3 method i.e should this only dispatch if x is a CMDSTANMODEl (lol I don't think that is actually the class name).

vignettes/wwinference.Rmd
gvegayon: We now have a data.frame method ;).

NEWS.md
gvegayon: Got it! That said, I think the NEWS.md file is improperly structured. I will put it on the bottom and then make an issue pointing this out (R CMD check highlights this).

vignettes/wwinference.Rmd
gvegayon: Missed modifying that! Will reverse the change.

gvegayon:
> @gvegayon A few suggested changes and some questions. Overall looking really good and excited to have this working! I think I am just trying to wrap my head around what the use case of the `.default` is

Here is a reprex illustrating the use of S3 methods with default:

``` r
myfun <- function(x) UseMethod("myfun")

myfun.data.frame <- function(x) {
  message("data.frame")
}

myfun.default <- function(x) {
  message("default")
}

myfun.integer <- function(x) {
  message("integer")
}

myfun(USArrests)
#> data.frame
```

``` r
myfun(1L)
#> integer
```

``` r
myfun("G")
#> default
```

<sup>Created on 2024-08-28 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

kaitejohnson (approved):
This LGTM. I presume we should punt reducing the warmup and sampling iterations to a separate PR?

kaitejohnson (approved):

kaitejohnson (approved):
I think GH won't let me merge this because I pushed to it last, @dylanhmorris would you mind approving additionally?

seabbs (approved):

PR #62, opened 2024-08-14
geom_line => geom_step

### LINKED ISSUE #60: Posterior predictive distributions should be piecewise constant?
image

In https://github.com/cdcent/cfa-spatial-ww-internal/issues/5#issue-2458143691 @cbernalz created these plots.

The red line is the data, and the other coloured lines I think are draws from the posterior predictive distribution. I think these draws from the posterior predictive distribution should be piecewise constant (like the data). This might be possible using e.g. `geom_step` (or otherwise not using `geom_line` to connect up the shortest path).

As I understand it the code for these plots was from somewhere in this package.

dylanhmorris:
Resolves #60. Also changes plot output to `svg` to avoid pixelation.

### DIFF
--- a/R/figures.R
+++ b/R/figures.R
@@ -167,9 +167,10 @@ get_plot_global_rt <- function(draws,
     draw %in% !!sampled_draws
   )
 
-  # R(t) of the hypothetical state
+  # R(t) timeseries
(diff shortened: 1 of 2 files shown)


kaitejohnson (approved):
LGTM!

PR #73, opened 2024-08-19
Minor formatting changes

akeyel:
Suggested typographical corrections from reading through the vignette


dylanhmorris (approved):
Thanks for catching these, @akeyel!

PR #74, opened 2024-08-22
42 parameterize weekday effect with a dirichlet

dylanhmorris:
This PR:

- Replaces the (constrained) Normal prior on the simplex underlying the day-of-week effects with a Dirichlet prior 
- Updates `model_definition.md` to reflect this.
- Specifies default parameters for the dirichlet prior in `example_params.toml` and `model_definition.md`. Parameters chosen to  make the Dirichlet prior similar to but less informative than the earlier constrained Normal.
- Tweaks the initialization function to ensure that entries to the initialization simplex will always be non-negative, but does not initialize with a random Dirichlet (defers that to the initialization refactor)
- Makes `to_simplex` error if given a vector with negative entries, and fixes typos in its documentation


seabbs:
Is there a reason not to declare the prior as a simplex? Also lots and lots of changes in this PR!

R/generate_simulated_data.R
```diff
@@ -143,11 +143,7 @@ generate_simulated_data <- function(r_in_weeks = # nolint
   model$expose_functions(global = TRUE)
 
   # Pull parameter values into memory
-  params <- get_params(input_params_path) # load in a single row tibble
-  par_names <- colnames(params) # pull them into memory
-  for (i in seq_along(par_names)) {
-    assign(par_names[i], as.double(params[i]))
-  }
+  params <- get_params(input_params_path)
```
seabbs, line 146: I can see why you maybe want to do this but it doesn't seem like it belongs in this PR

seabbs: For a future refactor I think its quite clunnky to get_params internally like this. I would either do it in the default settings of the function (i.e get_params(path)) or (much more likely) do it using fn args to give more idiomatic R code and then have a pattern where you use do.call externally if wanting to run against a param list

dylanhmorris:
> declare the prior as a simplex

Declare in what sense? Is this a point about the code or the model writeup?

R/generate_simulated_data.R
dylanhmorris: I agree but a minimal level of restructure was necessary so that vector-valued parameters ($\alpha$ vector of the Dirichlet prior) could be passed). And this accords with what is happening in #29

seabbs:
Sorry my mistake it already is a simplex

> simplex[7] hosp_wday_effect;

R/generate_simulated_data.R
seabbs: hmm okedokie

dylanhmorris:
> Sorry my mistake it already is a simplex (and has always been but with a weird prior)
> 
> > simplex[7] hosp_wday_effect;

Yes exactly this was the motivation for this change.

data-raw/covid_pmfs.R
```diff
@@ -7,20 +7,20 @@ params <- wwinference::get_params(
 
 # Put it all together
 generation_interval <- withr::with_seed(42, {
-  wwinference::simulate_double_censored_pmf(
+  wwinference:::simulate_double_censored_pmf(
```
kaitejohnson, line 10: @dylanhmorris Not sure what happened here but I think an extra `:` was added unintentionally?

inst/extdata/example_params.toml
```diff
@@ -39,8 +39,8 @@ p_hosp_w_sd_sd = 0.01
 
 inv_sqrt_phi_prior_mean = 0.1 # 1 / sqrt(100)
 inv_sqrt_phi_prior_sd = 0.1414214 # 1 / sqrt(50)
-wday_effect_prior_mean = 0.1428571 # 1 / 7
-wday_effect_prior_sd = 0.05
+hosp_wday_effect_prior_alpha = [5, 5, 5, 5, 5, 5, 5]
```
kaitejohnson, line 42: This might be an incredibly naive question, but can you explain your choice of all 5s here? Is this a recommended non-informative prior choice for a direchlet?

kaitejohnson (commented):
Just a quick question on the prior choice. Am guessing this is some sort of approximation of the number of trials similar to a beta prior?

data-raw/covid_pmfs.R
dylanhmorris: No this is necessary, since functions like `simulate_double_censored_pmf` not exported.

https://stackoverflow.com/questions/41582136/r-what-do-you-call-the-and-operators-and-how-do-they-differ

inst/extdata/example_params.toml
dylanhmorris: As I mentioned above, 
> Parameters chosen to make the Dirichlet prior similar to but less informative than the earlier constrained Normal.

And yes, for reasoning about the parameters you can use a similar [updating-based intuition](https://en.wikipedia.org/wiki/Dirichlet_distribution#Conjugate_to_categorical_or_multinomial) to the one for the Beta, just with more categories. So because all the numbers as the same (5), this prior will be centered on equal weekly reporting, just as the previous one was intended to be. But because none of the numbers are large, uneven reporting is a priori plausible. And of course, always good to check intuition via forward simulation (I did a bit of that as well).

kaitejohnson:
LGTM pending merge conflicts resolved from #29

kaitejohnson (dismissed):
LGTM pending merge conflicts resolved

R/generate_simulated_data.R
```diff
@@ -362,43 +345,51 @@ generate_simulated_data <- function(r_in_weeks = # nolint
...
-    # Calculate log concentration for the days that we have observations
-    log_obs_conc_lab_site[i, stl] <- log_obs_g_over_n_lab_site[i, stl] -
-      log(params$ml_of_ww_per_person_day)
-  }
+
+  # Global adjusted R(t) --------------------------------------------------
+  # I(t)/convolve(I(t), g(t)) #nolint
+  # This is not used directly, but we want to have it for comparing to the
+  # fit.
+  rt <- calc_rt(new_i_over_n,
```
kaitejohnson, line 387:
```suggestion
  rt <- calc_rt(
  new_i = new_i_over_n,
```
Caught my own error, would like to generally specify every function arg

kaitejohnson (commented):
Just one suggestion that is my own error from the previous PR. Otherwise looks good to me.

kaitejohnson (approved):
LGTM

PR #76, opened 2024-08-22
2024-08-22 update : preping testfile.

cbernalz:


R/wwinference.R
```diff
@@ -36,6 +36,8 @@
 #' `compile_model()`
 #' @param dist_matrix Distance matrix for spatial correlation in distance
 #' correlation function.
+#' @param bool_spatial_comp Boolean for whether or not user wants a spatial
+#' model.
```
kaitejohnson, line 40:
```suggestion
#' model. Default is `FALSE` indicating the spatial component is not being modeled
```

kaitejohnson (commented):
@cbernalz Can you provide a high-level description of what is being done in this PR? I'm a bit out of the loop since I've been out this week. 

High level question for you, is the "switch" here meant to be specific to the distance matrix being incorporated with some correlation function? Or is this a more general switch for whether or not we are inferring any sort of a correlation matrix (that may or may not use the distance matrix as an input), and then we will eventually add some functionality to specify the correlation matrix structure (that might use distance data or something else).

R/wwinference.R
```diff
@@ -71,11 +73,20 @@ wwinference <- function(ww_data,
...
+                        dist_matrix = NULL,
+                        bool_spatial_comp = FALSE) {
   # Check that data is compatible with specifications
   check_date(ww_data, model_spec$forecast_date)
   check_date(count_data, model_spec$forecast_date)
 
+  if (bool_spatial_comp == TRUE && is.null(dist_matrix)) {
+    stop(
+      "Spatial Components Desired, but Distance Matrix Not Supplied!!!\n
+          *distance matrix required for current implementation*"
```
kaitejohnson, line 85: Do we want this to be this kind of switch? Or do we want it to be a more general switch of whether we infer any sort of a correlation matrix?

R/wwinference.R
cbernalz: Currently this is a switch for if the user wants to use a spatial model based off distances or not.  If they do want to use spatial model based off distances there is a check to ensure they have provided a distance matrix.  This can be extended later when we introduce spatial model without distance matrix.

R/wwinference.R
kaitejohnson: Makes sense and think this error message should be here for now since we don't have the feature implemented currently.

R/wwinference.R
cbernalz: Do you mean that I should add another switch to which type of correlation (structured w/ dist_matrix or unstructured infer corr) and have default as structured w/ dist_matrix?  If so, I think we should wait on that for the next PR as the "frontend" will depend on how I plan to do this within the `wwinference.stan`.  But I think it will be exactly how you are suggesting.  Right now we only have the high level switch and I can improve the documentation of it.

dylanhmorris:
Before we land this, it would be good to pull in the latest changes from `main` @cbernalz. Could you fix the conflicts in #83 and request a review from me?

R/wwinference.R
kaitejohnson: That sounds reasonable to me, and exactly! Fine to do it the next PR

dylanhmorris (approved):
LGTM, thanks @cbernalz!

Issue #23, opened 2024-07-12
Set up caching of `wwinference.Rmd` vignette in R-cmd-check


kaitejohnson:
## Goal
Set up caching of vignette build. Right now, this takes a long time because it has to run the full example model.

## Requirements
- [ ] if nothing that impacts the vignette gets updated, we want to cache the vignette build

kaitejohnson:
OK, this might be a bit of a lift so other options are to set up some set cadence of updating like is done in epinowcast:
https://github.com/epinowcast/epinowcast/pull/494
https://github.com/epinowcast/epinowcast/pull/497

dylanhmorris:
Going to close this as resolved by #78 and open a new issue to discuss whether the website build needs speedup or is fine as is.

PR #78, opened 2024-08-23
Update r-cmd-check.yaml

dylanhmorris:
Since the mandatory website build CI will fail if vignettes cannot be built, we do not need to check them in the `R CMD check` action as well.

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -24,7 +24,7 @@ jobs:
       id-token: write
       pages: write
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - uses: r-lib/actions/setup-pandoc@v2
         with:
@@ -52,7 +52,7 @@ jobs:
         shell: Rscript {0}
 
       - name: Upload artifact for GH pages deployment
-        uses: actions/upload-pages-artifact@v2
+        uses: actions/upload-pages-artifact@v4
         with:
           path: "docs/"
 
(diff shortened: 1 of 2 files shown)


dylanhmorris:
This goes a long way toward resolving the problems discussed in #23, at least with `R CMD check` in CI. Whether we want to try to speed up website builds is another question.

kaitejohnson (approved):
This LGTM pending it passes. Can I ask a follow up question for my understanding. 

What do `--no-manual` and `--as-cran` ? mean and where did you go (am guessing the github repo for `actions/check-r-package` ?) to figure out which commands to put in `build_args` / `args`?

dylanhmorris:
> What do --no-manual and --as-cran ? mean and where did you go (am guessing the github repo for actions/check-r-package ?) to figure out which commands to put in build_args / args?

1. `--no-manual and --as-cran` are in there because they are the defaults for the action, so we were silently doing them before. I found that out by going to the action's github repo, as you guessed https://github.com/r-lib/actions/tree/v2-branch/check-r-package
2. if you are interested in what each flag does and what flags are available, a useful thing to do is `R CMD build --help` and `R CMD check --help` at the command line. It was by doing this that I worked out `no-vignettes` for the flag I wanted for check
3. But also some trial and error was important. For example, I first tried just setting the `args` for the action, but on watching the action run, I saw that that was not affecting the build step. I then noticed that there was a `build_args` option in the documentation, and used `R CMD build --help` to work out which flag to use.

PR #81, opened 2024-08-23
Fix params docs in wwinference.R

dylanhmorris:
params are now a list


kaitejohnson (dismissed):
LGTM but probably should rerun `devtools::document()` locally to update the corresponding doc?

kaitejohnson (approved):
LGTM

PR #82, opened 2024-08-23
Update `spatial-main` from `main`

dylanhmorris:
Merge updates to `main` into `spatial-main`


dylanhmorris:
Created in the wrong direction. Closing in favor of #83.

PR #85, opened 2024-08-23
Switch to placing prior on and inferring `i/n` at the first observed timepoint

### LINKED ISSUE #84: place a prior on and infer the last point in the exponential growth period, not the first
This is easier to reason about.

dylanhmorris:
Instead of the first unobserved timepoint. This should increase interpretability.

Closes #84


inst/extdata/example_params.toml
```diff
@@ -7,14 +7,17 @@ r_prior_sd = 1
 sigma_rt_prior = 0.1
 dur_inf = 7
 
-sigma_i0_prior_mode = 0
-sigma_i0_prior_sd = 0.5
-i0_certainty = 5
+sigma_i_first_obs_prior_mode = 0
```
kaitejohnson, line 10: Just a general comment that this `i_first_obs` is more explicit, but I wonder if we could just keep the `i0` as its a bit less verbose, and then make sure we are more clear in the model definition the `i0` or $I(t=0)$ is the start of the observations?

inst/stan/wwinference.stan
```diff
@@ -187,9 +190,9 @@ transformed parameters {
   s = get_vl_trajectory(t_peak, viral_peak, dur_shed, gt_max);
 
   // Site level disease dynamic estimates!
-  i0_site_over_n = inv_logit(logit(i0_over_n) + eta_i0 * sigma_i0);
-  growth_site = initial_growth + eta_growth * sigma_growth; // site level growth rate
   for (i in 1:n_subpops) {
+    vector[n_weeks] log_r_site_t_in_weeks;
+    real log_i0_site = log(i_first_obs_over_n_site[i]) - uot * initial_exp_growth_rate_site[i];
```
kaitejohnson, line 195: My suggestion around using `i0` to indicate the first observed time would of course require renaming this to `log_i_uot_site` or something that indicates its the log(I/N) at `t = - uot`

inst/stan/wwinference.stan
```diff
@@ -168,10 +169,12 @@ transformed parameters {
   vector[ot + uot + ht] state_inf_per_capita = rep_vector(0, uot + ot + ht); // state level incident infections per capita
   matrix[n_subpops, ot + ht] model_log_v_ot; // expected observed viral genomes/mL at all observed and forecasted times
   real<lower=0> g = pow(log10_g, 10); // Estimated genomes shed per infected individual
-  real<lower=0> i0 = i0_over_n * state_pop; // Initial absolute infection incidence
-  vector[n_subpops] i0_site_over_n; // site-level initial
-  // per capita infection incidence
-  vector[n_subpops] growth_site;
+  vector<lower=0, upper=1>[n_subpops] i_first_obs_over_n_site = inv_logit(logit(i_first_obs_over_n) +
```
kaitejohnson, line 172: Slight slight personal preference when reading stan code that the defining of variables, unless  extremely straightforward, just defines what they are and then the actual transformation happens sequentially.

So instead of this happening here, have it defined here and then have it happening right outside the loop

inst/stan/wwinference.stan
```diff
@@ -187,9 +190,9 @@ transformed parameters {
   s = get_vl_trajectory(t_peak, viral_peak, dur_shed, gt_max);
 
   // Site level disease dynamic estimates!
-  i0_site_over_n = inv_logit(logit(i0_over_n) + eta_i0 * sigma_i0);
-  growth_site = initial_growth + eta_growth * sigma_growth; // site level growth rate
   for (i in 1:n_subpops) {
+    vector[n_weeks] log_r_site_t_in_weeks;
```
kaitejohnson, line 194: Again, slight preference for all variables, even ones that are getting dynamically updated like this one, to be defined in the beginning of the block

kaitejohnson (commented):
Thank you so much for doing this @dylanhmorris! I think this is definitely more intuitive. See though my suggestion about continuing to use the `i0` notation. Curious your thoughts on that. 

Also minor notes on preference in stan file syntax. 

Also we should make sure to change the model definition alongside this. I think it was just ambiguous previously. I can branch off of this and try changing it to reflect the new parameterization if that would be helpful? 

Can you describe a bit what the major change is in the test of the `ihr_transform` and `rt_assmebly` -- is it just that we want to use the `get_ind_m()` function instead of the one passed into stan (that upstream called that function) so that we are more explicitly testing that function?

inst/stan/wwinference.stan
damonbayer: We have used the t=0 as time of first observation convention in pyrenew. Would be nice to be consistent.

inst/stan/wwinference.stan
seabbs: FYI this has an impact on memory as anything stored outside of brackets (even just empty brackets) is stored and anything inside isn't. This can also have a slight performance impact.

inst/stan/wwinference.stan
seabbs: @damonbayer that would suggest rewriting all the indexing in this model which doesn't seem worthwhile. I would suggest you have one approach per model and note down how to map between

inst/stan/wwinference.stan
dylanhmorris: Related: I think defining throwaway loop variables within the loop makes code slightly clearer. It indicates to the reader that those variables are throwaways and won't be reused/stored.

inst/stan/wwinference.stan
dylanhmorris: In the past, I've sometimes defined a helper function:

```Stan
vector center(real loc, real scale, vector deviations) {
   return loc + scale * deviations;
}
```
and then
```Stan
  vector[n_subpops] initial_exp_growth_rate_site = center(
      mean_initial_exp_growth_rate,
      sigma_initial_exp_growth_rate,
      eta_initial_exp_growth_rate);
```

kaitejohnson: I think I pretty much only do that if it is just a scalar transformation. Happy to add the function though, but I'd still prefer the operation be called right before the for loop where it is used.

inst/stan/wwinference.stan
kaitejohnson: @seabbs I'm not sure I understand your comment. Do you mean that this isn't stored because it's inside brackets (in this case a for loop?). 

I am fine with that @dylanhmorris I think that makes sense

seabbs: > Do you mean that this isn't stored because it's inside brackets (in this case a for loop?).

Yes it makes it a temporary variable that being said I wouldn't redefine it inside a loop like this as it will an a small allocation overhead each time. Instead if I felt worried I would do { int temp; for (.... temp = 1, use temp)}

dylanhmorris: The compiler reallocates? That's surprising to me. I'd have expected it to work out that it could allocate only on the first iteration and then overwrite.

kaitejohnson:
> > Thank you so much for doing this @dylanhmorris! I think this is definitely more intuitive. See though my suggestion about continuing to use the i0 notation. Curious your thoughts on that.
> 
> I think as long as we're clear about what we mean by `t=0` it's fine either way. I went with `first_obs` for backward compatibility and to avoid breaking anything accidentally. If we do want to change time indexing I think that's a separate PR.

Call me crazy, but I think we can still specify this as `i0` from the user define parameters and the variable names, when under the hood the time indexing starts at `-uot`. I see why you switch to `first_obs` though. I can do it in a separate PR, but I just find it more readable as `i0`

dylanhmorris:
> i0 from the user define parameters and the variable names, when under the hood the time indexing starts at -uot. I see why you switch to first_obs though. I can do it in a separate PR, but I just find it more readable as i0

Yes, absolutely, but it should be a separate PR imo.

inst/stan/wwinference.stan
dylanhmorris: See discussion below. @kaitejohnson is interested in changing notation to make `t=0` correspond to the first observation, but if that is to happen it should be a separate issue+PR.

dylanhmorris:
> Weirdly CI isn't telling me what failed..

It's the snapshot test (`test_ww_model.R`). I go back and forth on whether to improve it or abandon it, but either decision is beyond the scope of this PR imo.

kaitejohnson:
> > Weirdly CI isn't telling me what failed..
> 
> It's the snapshot test (`test_ww_model.R`). I go back and forth on whether to improve it or abandon it, but either decision is beyond the scope of this PR imo.

I think the thing that might have broke it was using the `generate_initial_values = TRUE`... again not ideal to have that off but I think the draws should be deterministic without that

inst/stan/wwinference.stan
kaitejohnson: And wouldn't change the actual time indexing, just how it is notated

dylanhmorris:
> I think the thing that might have broke it was using the generate_initial_values = TRUE... again not ideal to have that off but I think the draws should be deterministic without that if the stan seed is set

That was actually one thing I changed. It now generates initial values but fixes a seed. I strongly suspect the issue was architectural differences between the machine generating the snapshot and the machine running in CI, since I was able to resolve it by generating the snapshot somewhere else. This is the same thing @cbernalz ran into a while back. Again, it suggests a need to rethink the test a bit.

kaitejohnson (approved):
This LGTM, thanks again for doing this. I will open a separate issue to rename to I0 throughout out of syntax preference.

PR #86, opened 2024-08-25
Change output names

kaitejohnson:
@gvegayon This makes the requested changes to the output names as discussed here https://github.com/CDCgov/ww-inference-model/pull/58#discussion_r1730411415. The issue with the `fit` not being recognized as  CmdStanMCMC options is still persisting. I wonder if this should be `fit$result`? Didn't test but can!


gvegayon:
R CMD check failed because I incorrectly asserted the fit object during the `wwinference_fit` construction. Now fixed.

kaitejohnson:
Will let you merge!

PR #108, opened 2024-08-27
Addressing R CMD check notes due to tidyeval syntax

gvegayon:
- ~Adds `+ file LICENSE` to `DESCRIPTION`~ Adds `LICENSE` to the `.Rbuildignore` file (R convention).
- Moves `withr` to `Suggests`, as it is only used during testing.
- Adds the `rlang::.data` imports to address masking notes.
- Makes use of `.data$` for `dplyr` verbs when needed (see [here](https://dplyr.tidyverse.org/articles/in-packages.html)).
- Replaces expressions with strings in `dplyr` verbs when needed.
- ~Flags various variables to avoid warnings during R CMD check via `utils::globalVariables`. All variables in the list are those used in `tidybayes::spread_draws` calls (which uses expressions).~
- Adds some files/folders to `.Rbuildignore`.
- Modifies the `NEWS.md` document to follow R's convention (see also #118). 
- Ensures all `ggplot2` functions are imported and thus removes the use of `ggplot2::` (**NOTE: `dplyr::` is still heavily used).**


kaitejohnson (commented):
One how did you do this all in a day. You rock. 

Two I will admit I hate this so much... I know the other option is `.data[variable]` and that's not much better...

@seabbs suggested that there is a way to use `autoglobals()` and we could also import the common functions to avoid having to use things like `ggplot2::ggplot()` and such everywhere (though I know there is currently lots of redundancy in this). I just would prefer the code be a bit more readable...

gvegayon:
> One how did you do this all in a day. You rock.

Thank you!

> Two I will admit I hate this so much... I know the other option is `.data[variable]` and that's not much better...
> 
> @seabbs suggested that there is a way to use `autoglobals()` and we could also import the common functions to avoid having to use things like `ggplot2::ggplot()` and such everywhere (though I know there is currently lots of redundancy in this). I just would prefer the code be a bit more readable...

Yes, I don't like it either, but it's the way `dplyr` says it should be done. Just checked the [autoglobals alternative using roxygen tags](https://github.com/anthonynorth/roxyglobals); looks handy! One question: How reliable is that R package? We wouldn't be adding a dependency for the users, but we would be doing so for the developers. Perhaps avoid the package and use `utils::globalVariables()`? It would be pretty much the same thing!

dylanhmorris:
> * Flags various variables to avoid warnings during R CMD check via `utils::globalVariables`. All variables in the list are those used in `tidybayes::spread_draws` calls (which uses expressions).

Thanks for this, @gvegayon 

For `tidybayes::spread_draws`, we can do calls of the form

```R
fit |> tidybayes::spread_draws(!!str2lang("variable name"))
```

See https://github.com/mjskay/tidybayes/issues/38#issuecomment-895281257

R/generate_simulated_data.R
```diff
@@ -418,7 +418,7 @@ generate_simulated_data <- function(r_in_weeks = # nolint
   ) |>
     dplyr::rename(
       daily_hosp_admits_for_eval =
-        daily_hosp_admits
+        .data$daily_hosp_admits
```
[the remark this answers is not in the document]
dylanhmorris: For consistency with how you're doing `rename` calls elsewhere

R/get_stan_data.R
```diff
@@ -239,9 +239,9 @@ get_stan_data <- function(input_count_data,
   # Get the total pop, coming from the larger population generating the
   # count data
   pop <- input_count_data |>
-    dplyr::select(total_pop) |>
+    dplyr::select("total_pop") |>
     unique() |>
```
dylanhmorris, line 243:
```suggestion
```

R/get_stan_data.R
```diff
@@ -239,9 +239,9 @@ get_stan_data <- function(input_count_data,
...
-    dplyr::pull(total_pop)
+    dplyr::pull(.data$total_pop)
```
dylanhmorris, line 244:
```suggestion
```

R/model_component_fwd_sim.R
```diff
@@ -371,18 +371,21 @@ format_ww_data <- function(log_obs_conc_lab_site,
     dplyr::mutate(
       lod_sewage =
```
dylanhmorris, line 372:
```suggestion
      "lod_sewage" =
```

R/preprocessing.R
```diff
@@ -52,15 +52,15 @@ preprocess_ww_data <- function(ww_data,
   ww_data_add_cols <- ww_data |>
     dplyr::left_join(
       ww_data |>
-        dplyr::distinct(lab, site) |>
+        dplyr::distinct(.data$lab, .data$site) |>
         dplyr::mutate(
           lab_site_index = dplyr::row_number()
```
dylanhmorris, line 57:
```suggestion
         "lab_site_index" = dplyr::row_number()
```

R/preprocessing.R
```diff
@@ -52,15 +52,15 @@ preprocess_ww_data <- function(ww_data,
...
         ),
       by = c("lab", "site")
     ) |>
     dplyr::left_join(
       ww_data |>
-        dplyr::distinct(site) |>
+        dplyr::distinct(.data$site) |>
         dplyr::mutate(site_index = dplyr::row_number()),
```
dylanhmorris, line 64:
```suggestion
        dplyr::mutate("site_index" = dplyr::row_number()),
```

R/preprocessing.R
```diff
@@ -178,20 +178,20 @@ flag_ww_outliers <- function(ww_data,
     # exclude below LOD from z scoring and remove lab-sites with too
     # few data points
     dplyr::filter(
-      below_lod == 0,
-      n_data_points > threshold_n_dps
+      .data$below_lod == 0,
+      .data$n_data_points > threshold_n_dps
```
dylanhmorris, line 182:
```suggestion
      .data$n_data_points > !!threshold_n_dps
```

R/preprocessing.R
```diff
@@ -202,40 +202,41 @@ flag_ww_outliers <- function(ww_data,
...
       ww_rho |>
-        dplyr::group_by(lab_site_index) |>
+        dplyr::group_by(.data$lab_site_index) |>
         dplyr::summarise(
-          mean_rho = mean(rho, na.rm = TRUE),
-          std_rho = sd(rho, na.rm = TRUE),
+          mean_rho = mean(.data$rho, na.rm = TRUE),
+          std_rho = sd(.data$rho, na.rm = TRUE),
           mean_conc = mean(!!rlang::sym(conc_col_name), na.rm = TRUE),
           std_conc = sd(!!rlang::sym(conc_col_name), na.rm = TRUE)
```
dylanhmorris, line 210:
```suggestion
          mean_conc = mean(.data[[conc_col_name]], na.rm = TRUE),
          std_conc = sd(.data[[conc_col_name]], na.rm = TRUE)
```

R/preprocessing.R
```diff
@@ -202,40 +202,41 @@ flag_ww_outliers <- function(ww_data,
...
         ),
       by = "lab_site_index"
     ) |>
-    dplyr::group_by(lab_site_index) |>
-    mutate(
-      z_score_conc = (!!rlang::sym(conc_col_name) - mean_conc) / std_conc,
-      z_score_rho = (rho - mean_rho) / std_rho
+    dplyr::group_by(.data$lab_site_index) |>
+    dplyr::mutate(
+      z_score_conc = (!!rlang::sym(conc_col_name) - .data$mean_conc) /
```
dylanhmorris, line 216:
```suggestion
      z_score_conc = (.data[[conc_col_name]] - .data$mean_conc) /
```

R/preprocessing.R
```diff
@@ -202,40 +202,41 @@ flag_ww_outliers <- function(ww_data,
...
+        is.na(.data$z_score_conc) ~ 0,
         TRUE ~ 0
       ),
       flagged_for_removal_rho = dplyr::case_when(
         (
-          abs(z_score_rho) >= rho_threshold &
-            (abs(z_score_rho_t_plus_1) >= rho_threshold) &
-            sign(z_score_rho != sign(z_score_rho_t_plus_1))
+          abs(.data$z_score_rho) >= rho_threshold &
+            (abs(.data$z_score_rho_t_plus_1) >= rho_threshold) &
```
dylanhmorris, line 230:
```suggestion
            (abs(.data$z_score_rho_t_plus_1) >= !!rho_threshold) &
```

R/preprocessing.R
```diff
@@ -202,40 +202,41 @@ flag_ww_outliers <- function(ww_data,
...
+        abs(.data$z_score_conc) >= log_conc_threshold ~ 1,
+        is.na(.data$z_score_conc) ~ 0,
         TRUE ~ 0
       ),
       flagged_for_removal_rho = dplyr::case_when(
         (
-          abs(z_score_rho) >= rho_threshold &
-            (abs(z_score_rho_t_plus_1) >= rho_threshold) &
-            sign(z_score_rho != sign(z_score_rho_t_plus_1))
+          abs(.data$z_score_rho) >= rho_threshold &
```
dylanhmorris, line 229:
```suggestion
          abs(.data$z_score_rho) >= !!rho_threshold &
```

R/preprocessing.R
```diff
@@ -202,40 +202,41 @@ flag_ww_outliers <- function(ww_data,
...
+        .data$std_conc,
+      z_score_rho = (.data$rho - .data$mean_rho) / .data$std_rho
     ) |>
     dplyr::mutate(
-      z_score_rho_t_plus_1 = dplyr::lead(z_score_rho, 1),
+      z_score_rho_t_plus_1 = dplyr::lead(.data$z_score_rho, 1),
       flagged_for_removal_conc = dplyr::case_when(
-        abs(z_score_conc) >= log_conc_threshold ~ 1,
-        is.na(z_score_conc) ~ 0,
+        abs(.data$z_score_conc) >= log_conc_threshold ~ 1,
```
dylanhmorris, line 223:
```suggestion
        abs(.data$z_score_conc) >= !!log_conc_threshold ~ 1,
```

dylanhmorris (commented):
Thanks so much for this, @gvegayon. A number of things, all very small.

kaitejohnson (commented):
Can we maybe have a quick chat before this gets merged in? I think I need to spend some time looking into the `autoglobals()` package/ what @gvegayon suggested with `utils::globalvariables` and understand the tradeoffs of that approach (which seem to just be a developer dependency?) over the `dplyr()` recommendation... 

I have a pretty strong preference to avoid resyntaxing the whole package this way. I know it doesn't materially change anything, but I personally find it super hard to follow and I think there is a chance that this deters users trying to understand what the code is doing. And since this is an R package for a very specific use case of a model rather than as say a framework for model development, I'm inclined to take some short cuts on optimal development, at least in this first release, to make it more approachable to people familiar with R/tidyverse syntax.

seabbs:
This is a lot of work so well done @gvegayon but as discussed here I find this code nearly unreadable and the benefits seem very small. 

I think if we used the package namespace and imported our core tooling (i.e from dplyr, and ggplot2) this would help a lot with the `package::function` bloat. I see few if any downsides to doing that. 

The downside to not using `.data$var` is that as @gvegayon points out it is recommended practice and not doing so could add some small danger. However I think the risk from this is fairly marginal in most of this code base so would prefer we using globals (but I feel less strongly about this than the package::function who's removal would make the code much easier to read and so make this issue less). 

In terms of just manually assigning the globals. Yes, we could but it is exceptionally tedious. As it's a light dev dependency where any change will run through CI I don't think having it as a dev dependency is a serious issue. @athowes how have you found it?

athowes:
I don't know that I understand the full context here but I've found using the [`roxyglobals`](https://github.com/anthonynorth/roxyglobals) easy enough.

About the other parts, it's not really my place to say / it's up to what you collectively decide, but if I were developing this package I would probably have reservations about the `pkg::` to everything and `.data$var` points. Perhaps that's just selfishness (it makes development less nice) but I think there are reasonable arguments to be made about code readability reducing the amount of error. With something like putting `ggplot2::` I might also question the extent to which this is ever likely to create bugs that aren't obvious / very easy to fix (I think that with some other more internal functions you might expect that an undetected namespace clash has more of a risk of creating a hard to see bug).

kaitejohnson:
@gvegayon I'm inclined to give the `roxyglobals` a try (happy to do this in a separate PR and then compare the pros and cons). I think I saw that you had already started adding `@autoglobal` to some of the new functions you wrote?

kaitejohnson:
Ok after discussing face-to-face, it seems like this is going to have to be the way to go in terms of the `.data[[]` and string calls, so will defer to @dylanhmorris on the requested changes. 

One remaining question I have is whether we could still do some amount of importing common package functions like `ggplot`, `aes`, `geom_line`, etc. or does that also always cause errors? It doesn't matter too much but I think it might help slightly with readability and is definitely hard to get into the habit of doing for things like ggplot

dylanhmorris:
Yes, important to separate the question of importing any/many functions or other names from other packages from the question of how to handle column names in tidyeval functions. We can continue to import things like `ggplot2::ggplot` etc (and note that per my review this PR sticks with importing `ggplot2::aes`)

dylanhmorris:
> .data[[] and string calls
Clarification that there are some places we'd be able to get away with tidy syntax but other places (dynamic column names) where we'll definitely need the `.data` pronoun: 

https://github.com/CDCgov/ww-inference-model/blob/7bcbebc4965b52759f9ba7b80c71f67f5b6cbc7b/R/preprocessing.R#L299

But given that:
- we'll be using `.data` of necessity in some places
- given that it's the tidyverse recommendation for packaged code as @gvegayon notes

I favor just using it throughout

dylanhmorris:
In short: I support removing the namespace qualifications from this PR (e.g. revert `ggplot2::ggplot()` to `ggplot()`, but keeping the universal use of the `.data` pronoun. @kaitejohnson, @gvegayon et al does that sound reasonable?

kaitejohnson:
> In short: I support removing the namespace qualifications from this PR (e.g. revert `ggplot2::ggplot()` to `ggplot()`), but keeping the universal use of the `.data` pronoun. @kaitejohnson, @gvegayon et al does that sound reasonable?

Yes I agree. I would also support importing a few other ggplot functions like `geom_line()`, `geom_point()`, `geom_vline/hline()` etc. but again this isn't super critical because I don't intend this package will be doing a ton of plotting anyway.

NAMESPACE
```diff
@@ -71,6 +71,7 @@ importFrom(ggplot2,theme)
 importFrom(lubridate,ymd)
 importFrom(posterior,as_draws_list)
 importFrom(posterior,subset_draws)
+importFrom(rlang,.data)
```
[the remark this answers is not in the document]
gvegayon: I think we can!

gvegayon:
Thanks all for the comments. I will add `ggplot2`'s function to imports and keep the `.data` syntax.

R/wwinference.R
```diff
@@ -121,7 +121,10 @@ wwinference <- function(ww_data,
   if (generate_initial_values) {
     init_lists <- c()
     for (i in 1:mcmc_options$n_chains) {
-      init_lists[[i]] <- get_inits_for_one_chain(stan_args, params)
+      init_lists[[i]] <- get_inits_for_one_chain(
```
gvegayon, line 124: @kaitejohnson, I think this fixes a bug. Before, it was getting `params` from somewhere I was not sure. Can you validate that what I did is correct?

R/generate_simulated_data.R
```diff
@@ -184,8 +184,8 @@ generate_simulated_data <- function(r_in_weeks = # nolint
   )
 
   forecast_date <- date_df |>
-    dplyr::filter(t == ot + nt) |>
-    dplyr::pull(date)
+    dplyr::filter(.data$t == ot + nt) |>
```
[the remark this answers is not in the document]
gvegayon: I am not following what the need of the double bang operator is here.

R/generate_simulated_data.R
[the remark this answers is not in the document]
gvegayon: Good point, fixing.

R/wwinference.R
kaitejohnson: Oh yup that was a bug that probably hasn't gotten caught in the vignette because we defined `params` in the vignette.... 

Also realizing this is out of date we need to merge #58!

R/get_draws_df.R
```diff
@@ -61,45 +79,45 @@ get_draws_df <- function(ww_data,
       flag_as_ww_outlier = NA,
       exclude = NA
```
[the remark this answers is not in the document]
gvegayon: All mutates now use `nameofvariable = ` instead of `"nameofvariable"`.

R/get_stan_data.R
```diff
@@ -68,11 +68,11 @@ get_input_ww_data_for_stan <- function(preprocessed_ww_data,
   # data. Arrange data for indexing. This is what will be returned.
   ww_data <- preprocessed_ww_data |>
     dplyr::filter(
-      exclude != 1,
-      date > last_count_data_date -
+      .data$exclude != 1,
+      .data$date > last_count_data_date -
```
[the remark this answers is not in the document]
gvegayon: Is this really needed?

R/get_stan_data.R
```diff
@@ -68,11 +68,11 @@ get_input_ww_data_for_stan <- function(preprocessed_ww_data,
...
         lubridate::days(calibration_time)
```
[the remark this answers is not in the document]
gvegayon: Same here

R/get_stan_data.R
```diff
@@ -239,9 +239,9 @@ get_stan_data <- function(input_count_data,
   # Get the total pop, coming from the larger population generating the
   # count data
   pop <- input_count_data |>
-    dplyr::select(total_pop) |>
+    dplyr::select("total_pop") |>
```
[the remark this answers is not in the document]
gvegayon: This seems like a big change; am I missing something? cc @kaitejohnson

R/get_draws_df.R
```diff
@@ -56,7 +56,7 @@ get_draws_df <- function(ww_data,
   draws <- fit_obj$result$draws()
 
   count_draws <- draws |>
-    tidybayes::spread_draws(pred_hosp[t]) |>
+    tidybayes::spread_draws(!!str2lang("pred_hosp[t]")) |>
```
gvegayon, line 59: @dylanhmorris, this change breaks `pkgdown` but not `R CMD check`. Any idea why?

R/get_stan_data.R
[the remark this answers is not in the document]
kaitejohnson: No so basically I just want the unique values of the total population column in the input count data here. 

I tend to do this with `select` for the column name and then `pull()` to get the vector, and then `unique()`. I often forget that this can be done more succinctly with `distinct(col_name)` and `pull()` bc distinct automatically does the selecting.

R/get_draws_df.R
```diff
@@ -113,17 +131,17 @@ get_draws_df <- function(ww_data,
       flag_as_ww_outlier = NA,
       exclude = NA
     ) |>
-    dplyr::select(-t)
+    dplyr::select(-"t")
 
   site_level_rt_draws <- draws |>
-    tidybayes::spread_draws(r_site_t[site_index, t]) |>
-    dplyr::rename(pred_value = r_site_t) |>
+    tidybayes::spread_draws(!str2lang("r_site_t[site_index, t]")) |>
```
dylanhmorris, line 137: Just a missing exclamation point @gvegayon 
```suggestion
    tidybayes::spread_draws(!!str2lang("r_site_t[site_index, t]")) |>
```

R/get_draws_df.R
dylanhmorris: Based on the logs, it looks like it's not the one you highlighted, it's this one: https://github.com/CDCgov/ww-inference-model/pull/108/files#r1737876278

so I think it's just the missing exclamation point.

seabbs:
> .data[[] and string calls


It would be best if having f2f calls to put the resulting conversation/justifications explicitly in the PR notes or it can make decision-making hard to track. 

I assume there you mean their recommendation to use `.data$var` and not `.data[[var]]` which they don't make a strong case for that I can see.  From the discussion that is here it seems a bit that those things might be conflated (i.e if we are really following `dplyr` standards we should be supporting NSE as the tool for specifying input in our functions.

kaitejohnson:
Here's how I would summarize the discussion: 

- tidyeval/`dplyr` error prone because of ambiguity with column names versus variables being passed to functions.
- R packages can look ugly/difficult to read under the hood. I personally am strongly in favor of readability, especially at this early stage, for those that would want to use this as I think it encourages contributions and makes it more approachable for someone to understand what a function is doing. 
- Using `.data[[var]]` whenever a column is being assigned, and using `.data$var` when something is being referenced, and using `.data[["var_name"]] whenever calling on a specific column name, and the `!!` whenever something is being assigned that is a variable not a column name....  (@gvegayon @dylanhmorris please correct me if I'm wrong/perhaps in this PR we could just clarify that so we know going forward what the standards are) 

My takeaway:
- package should have never been built using `dplyr` from the start given this. However, this would have made development a bit slower (for me personally) since despite being a modeling package, there is still a fair bit of data wrangling that needs to happen to format things properly with the two datasets

seabbs:
>  error prone because of ambiguity with column names versus variables being passed to functions.

Is there evidence for this assertion? 

> R packages can look ugly/difficult to read under the hood

They can also not right.

> package should have never been built using dplyr f

Interestingly we will be switching `epidist` over from `data.table` to `dplyr` as it makes it much easier to get contributions. 

> Using

Sounds like a clear plan. Should add this steer to the contributing guide.

gvegayon:
>> error prone because of ambiguity with column names versus variables being passed to functions.
>
> Is there evidence for this assertion?

The first thing I can think of is this issue with masking/unmasking caught by CFA in the Rt pipeline: [Teams discussion](https://teams.microsoft.com/l/message/19:63572bdbd6ec4f6080d3c103fa2ba2f5@thread.tacv2/1707942749024?tenantId=9ce70869-60db-44fd-abe8-d2767077fc8f&groupId=4f34eba7-8fcb-4a26-9b34-a434ea777f0c&parentMessageId=1707942749024&teamName=CFA-Predict&channelName=Topic%20-%20Dev%20-%20R&createdTime=1707942749024). I was able to reproduce the bug with had to do with NSE [here](https://github.com/cdcent/cfa-predict-sandbox/blob/2efa71ffb6e867bae7f061c12afd26223774316c/vegayon-xrd4/2024-02-14-scoping.md)

R/generate_simulated_data.R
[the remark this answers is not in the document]
dylanhmorris: `ot` and `nt` are variables, so they should be evaluated. There _shouldn't_ be clashing column names but should is a dangerous word, so I always like to have variables on the RHS of `tidyeval` expressions evalated explicitly

dylanhmorris:
> Using .data[[var]] whenever a column is being assigned, and using .data$var when something is being referenced, and using .data[["var_name"]] whenever calling on a specific column name, and the !! whenever something is being assigned that is a variable not a column name.... (@gvegayon @dylanhmorris please correct me if I'm wrong/perhaps in this PR we could just clarify that so we know going forward what the standards are)

This is close but not exact. As @seabbs says it may be simplest to follow the conventions in the `dplyr` vignette, particularly
https://dplyr.tidyverse.org/articles/in-packages.html#data-masking-and-tidy-selection-notes

There are a few places it is silent, however, and we can and should just choose a convention and be consistent

R/figures.R
```diff
@@ -33,39 +33,40 @@ get_plot_forecasted_counts <- function(draws,
...
+    ggplot2::geom_point(
       data = count_data_eval,
-      aes(x = date, y = .data[[count_data_eval_col_name]]),
+      aes(x = .data$date, y = .data[[count_data_eval_col_name]]),
       shape = 21, color = "black", fill = "white"
     ) +
-    geom_point(aes(x = date, y = observed_value)) +
-    geom_vline(aes(xintercept = lubridate::ymd(forecast_date)),
+    ggplot2::geom_point(aes(x = .data$date, y = .data$observed_value)) +
+    ggplot2::geom_vline(aes(xintercept = lubridate::ymd(forecast_date)),
```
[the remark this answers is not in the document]
gvegayon: Addressed in 7b24e012982136dc989699cde5bf3db7d64a969a

R/get_stan_data.R
gvegayon, line 1: Adding these changes broke R CMD check, @dylanhmorris

kaitejohnson:
> Interestingly we will be switching epidist over from data.table to dplyr as it makes it much easier to get contributions.

I think that makes sense. Curious if you all use `autoglobals()` or follow the `dplyr` guidelines? 

This also points out that we need to write up contributing guidelines...

R/get_stan_data.R
dylanhmorris: I think it's the `distinct |> pull()`, not the `!!`

```R
Error in `sum(ww_values$pop_ww)/pop`: non-numeric argument to binary operator
```

dylanhmorris: Learned something new! The issue is that `distinct` isn't tidyselect under the hood.

```R
df <- tibble::tibble(a = c(1,2,3), b = c(1, 1, 2))

df |> dplyr::distinct("b")
```
```R
# A tibble: 1 × 1
  `"b"`
  <chr>
1 b
```

whereas

```R
df |> dplyr::distinct(data$b)
```
yields the expected result:
```R
# A tibble: 2 × 1
      b
  <dbl>
1     1
2     2
```

R/get_stan_data.R
```diff
@@ -239,9 +239,8 @@ get_stan_data <- function(input_count_data,
   # Get the total pop, coming from the larger population generating the
   # count data
   pop <- input_count_data |>
-    dplyr::select("total_pop") |>
-    unique() |>
-    dplyr::pull(.data$total_pop)
+    dplyr::distinct("total_pop") |>
```
dylanhmorris, line 242:
```suggestion
    dplyr::distinct(.data$total_pop) |>
```

dylanhmorris (commented):
Fixing the bug I introduced, apologies @gvegayon

R/get_draws_df.R
```diff
@@ -1,3 +1,21 @@
...
+#   c(
+#     "pred_hosp",
+#     "t",
+#     "pred_ww",
+#     "lab_site_index",
+#     "rt",
+#     "r_site_t",
+#     "site_index"
+#   )
+# )
```
dylanhmorris, line 16: This can now be deleted, no?

R/get_draws_df.R
```diff
@@ -1,3 +1,21 @@
...
+  c(
+    "pred_hosp",
+    "t",
+    "pred_ww",
+    "lab_site_index",
+    "rt",
+    "r_site_t",
+    "site_index"
+  )
+)
```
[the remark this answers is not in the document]
dylanhmorris: Remove these now that we've used the `sym2lang` trick?

seabbs:
> Curious if you all use autoglobals() or follow the dplyr guidelines?

Autoglobals for the left hand side and nse for the right as needed I think

kaitejohnson:
Btw since I presume this will be merged pretty close to as is and I expect it will lead to many git conflicts if I don't branch off if it, I am going to work from this PR. I would approve and merge but I see a number of unresolved conversations and I don't want to commit those suggestion myself. Looks to me like they are mostly choice of syntax.

R/generate_simulated_data.R
```diff
@@ -186,8 +186,8 @@ generate_simulated_data <- function(r_in_weeks = # nolint
   )
 
   forecast_date <- date_df |>
-    dplyr::filter(t == ot + nt) |>
-    dplyr::pull(date)
+    dplyr::filter(.data$t == !!ot + !!nt) |>
+    dplyr::pull(.data$date)
```
dylanhmorris, line 190: Not for now, but for future reference this can be `dplyr::pull("date")`

R/get_stan_data.R
```diff
@@ -14,7 +14,7 @@ get_input_count_data_for_stan <- function(preprocessed_count_data,
 
   input_count_data_filtered <- preprocessed_count_data |>
     dplyr::filter(
-      date > last_count_data_date - lubridate::days(calibration_time)
+      .data$date > last_count_data_date - lubridate::days(calibration_time)
```
dylanhmorris, line 17:
```suggestion
      .data$date > last_count_data_date - lubridate::days(!!calibration_time)
```

R/preprocessing.R
```diff
@@ -178,20 +178,20 @@ flag_ww_outliers <- function(ww_data,
...
-      n_data_points > threshold_n_dps
+      .data$below_lod == 0,
+      .data$n_data_points > !!threshold_n_dps
     ) |>
-    dplyr::group_by(lab_site_index) |>
-    dplyr::arrange(date, "desc") |>
+    dplyr::group_by(.data$lab_site_index) |>
+    dplyr::arrange(.data$date, "desc") |>
     dplyr::mutate(
       log_conc = log(!!rlang::sym(conc_col_name)),
```
dylanhmorris, line 187:
```suggestion
      log_conc = log(.data[[conc_col_name]]),
```

dylanhmorris (changes requested):
A couple very final things @kaitejohnson

R/generate_simulated_data.R
kaitejohnson: Can you explain why it should be here -- is it because it is a column name and we should always refer to column names by "col_name" if they are not variables for column names?

dylanhmorris: "Should" might be strong, but the reason it _can_ goes back to the distinction between "tidy selection" and "data masking". `dplyr::select()` and `dplyr::pull()`, among others, use[^1] tidy selection; functions like `dplyr::mutate()` and `dplyr::summarise()` use[^1] data-masking.

[^1]: For positional arguments after the first argument, which is typically the data frame to be manipulated

See https://dplyr.tidyverse.org/articles/programming.html#introduction

dylanhmorris: The result is that 

```R
df |> dplyr::select("a_string")
```
means "select the column named `'a_string'` if it exists"

whereas:

```R
df |> dplyr::mutate(new_col = "a_string")
```
is interpreted as "create a new column named `'new_col'` whose value in each row is the string `'a_string'`. 

And if what you want is "create a new column named `'new_col'` whose values are the values of another column named `'a_string'`, you need:

```R
df |> dplyr::mutate(new_col = .data[["a_string"]])
```

or 

```R
df |> dplyr::mutate(new_col = .data$a_string)
```

dylanhmorris (approved):
LGTM, thanks @gvegayon / @kaitejohnson!

R/generate_simulated_data.R
kaitejohnson: Or:
```r
df |> dplyr::mutate("new_col" = .data$a_string
```
Which I believe is the syntax we're rolling with here correct (we should always explicitly use the `""` for creating a new column name)?

PR #113, opened 2024-08-28
Fix ignored compute_likelihood bug

### LINKED ISSUE #112: `compute_likelihood` flag returned by `get_mcmc_options` is ignored
It is settable: 

https://github.com/CDCgov/ww-inference-model/blob/410857b4634d31df7585060082d9c2dc539b7bbf/R/wwinference.R#L245-L274

But `wwinference()` ignores it and sets `compute_likelihood = 1`

https://github.com/CDCgov/ww-inference-model/blob/410857b4634d31df7585060082d9c2dc539b7bbf/R/wwinference.R#L106-L118

I would advocate moving the flag out of `get_mcmc_options()` (since everything else there is passed directly as a keyword argument to `model$sample()`. I would think the place it would belong if settable is in `get_model_spec()` (alongside the `include_ww` flag.

dylanhmorris:
Closes #112

### DIFF
--- a/R/wwinference.R
+++ b/R/wwinference.R
@@ -113,8 +113,8 @@ wwinference <- function(ww_data,
     inf_to_count_delay = model_spec$inf_to_count_delay,
     infection_feedback_pmf = model_spec$infection_feedback_pmf,
     params = model_spec$params,
-    include_ww = as.numeric(model_spec$include_ww),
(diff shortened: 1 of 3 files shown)


kaitejohnson:
LGTM, thanks for catching this!

PR #114, opened 2024-08-28
Change names of PMFs

### LINKED ISSUE #111: Rename package data defined in `covid_pmfs.R`
Suggest renaming `generation_interval` to `default_covid_generation_interval` and `inf_to_hosp` to `default_covid_inf_to_hosp`

kaitejohnson:
This closes #111


R/wwinference.R
```diff
@@ -315,7 +315,7 @@ get_model_spec <- function(
     )) {
   model_specs <- list(
     generation_interval = generation_interval,
-    inf_to_count_delay = inf_to_hosp,
+    inf_to_count_delay = inf_to_count_delay,
```
dylanhmorris, line 318: Also looks like this may have fixed a bug in which the `inf_to_count_delay` wasn't actually configurable and instead always used the package data one. Is that your sense too? We should add a test for configurability (additional PR)

dylanhmorris (approved):
LGTM. One note about a potential out of scope new test.

R/wwinference.R
kaitejohnson: Yeah this is a good point. I think this worked because we assigned `inf_to_hosp` above anytime we called the function in the examples or tests. Will make an issue.

PR #117, opened 2024-08-28
Fix how hierarchical sd values are computed

dylanhmorris:


inst/stan/wwinference.stan
```diff
@@ -127,9 +127,9 @@ parameters {
   real<lower=0> sigma_growth;
   real<lower=-1, upper=1> initial_growth; // initial growth from I0 to first observed time
   real<lower=1/sqrt(5000)> inv_sqrt_phi_h;
-  real<lower=0> sigma_ww_site_mean; //mean of site level stdev
-  real<lower=0> sigma_ww_site_sd; // stdev of site level stdev
-  vector<lower=0>[n_ww_lab_sites]sigma_ww_site_raw; // let each lab-site combo have its own observation error
+  real mode_sigma_ww_site; //mode of site level stdev
+  real<lower=0> sd_log_sigma_ww_site; // stdev of the log site level stdev
```
kaitejohnson, line 131: Isn't this now a more wider prior because the prior values of the stdev are the same, but this is now in log scale?

inst/stan/wwinference.stan
```diff
@@ -290,9 +292,11 @@ model {
...
   initial_growth ~ normal(initial_growth_prior_mean, initial_growth_prior_sd);
   inv_sqrt_phi_h ~ normal(inv_sqrt_phi_prior_mean, inv_sqrt_phi_prior_sd);
-  sigma_ww_site_mean ~ normal(sigma_ww_site_prior_mean_mean, sigma_ww_site_prior_mean_sd);
-  sigma_ww_site_sd ~ normal(sigma_ww_site_prior_sd_mean, sigma_ww_site_prior_sd_sd);
-  sigma_ww_site_raw ~ std_normal();
+  mode_sigma_ww_site ~ normal(mode_sigma_ww_site_prior_mode,
+                              mode_sigma_ww_site_prior_sd);
+  sd_log_sigma_ww_site ~ normal(sd_log_sigma_ww_site_prior_mode,
+                                sd_log_sigma_ww_site_prior_sd);
+  eta_log_sigma_ww_site ~ std_normal();
```
kaitejohnson, line 299: This is a bit of a change in the actual implementation, correct?

You are now estimating the deviations on the log scale rather than natural scale?

kaitejohnson (commented):
Confirming my understanding that this changes the implementation, such that we are now assuming that the site level observation errors are lognormally distributed, rather than normally distributed around the mean (`sigma_ww_site)`?

inst/stan/wwinference.stan
dylanhmorris: Yes. Doing it on the natural scale will cause bounds problems

kaitejohnson: See my question above. Just want to make sure the priors are somewhat comparable even as we switch the form of the distribution?

dylanhmorris: Resolved face to face that for now we'll use `sd_log_sigma_ww_site_prior_sd` ~= log(2) and reconsider default priors later

kaitejohnson (approved):
Thanks for catching this @dylanhmorris. This makes sense to me

inst/stan/wwinference.stan
```diff
@@ -266,7 +266,9 @@ transformed parameters {
   // Option to add a population offset here at some point  log(model_V) + site_level_multiplier+ pop_ww[ww_sampled_sites]
 
   // Get the transformed lab-site level error (NCP for sigma_site ~ n(mean_sigma_site, sigma_sigma_ww_site))
```
seabbs, line 268: comment needs updating

seabbs:
Is this a fix to match to math outline, a fix to match previous understanding or a change in model? If its the latter does it also need an update in the math description of the model

seabbs:
Hmm having a look it looks like the model description doesn't really say which feels bad

kaitejohnson:
Yeah this unfortunately wasn’t properly documented anywhere and that’s on me, and is likely why the bug persisted for this long. I will write up an issue tomorrow to describe what was previously being done and to change the model definition to describe the current implementation. Should also add to the change log.

seabbs:
>  properly documented anywhere and that’s on me

That isn't how teams work

Issue #106, opened 2024-08-27
Do you need to set a starting seed for repeatability?


akeyel:
## Describe the improvement that needs to be made
How repeatable are the MCMC model runs? I.e. would you get different results if you ran the model multiple times?
Would using set.seed() be honored by the underlying STAN code, or would it just get ignored?

dylanhmorris:
The Stan PRNG seed is set separately from the R PRNG seed. Within the package, it is configured via `get_mcmc_options`
https://cdcgov.github.io/ww-inference-model/reference/get_mcmc_options.html

It strikes me that:

1. This should be better documented.
2. Currently, there is a fixed default (`seed = 123`). That should not be the case. I think I would advocate for default `seed = NULL`. 
3. There's also the matter of seeding the random initial values. See #107.

dylanhmorris:
This is addressed as of #175

PR #125, opened 2024-08-30
update vignette to reflect default NULL seed in mcmcoptions

### LINKED ISSUE #107: R PRNG seeding when `generate_random_inits = TRUE`?
https://github.com/CDCgov/ww-inference-model/blob/410857b4634d31df7585060082d9c2dc539b7bbf/R/wwinference.R#L121-L126

We should either build in functionality for optional seed setting as an argument to `wwinference` or explicitly direct the user to run `wwinference` in the context of an explicitly seeded PRNG if they want reproducible results. 

@kaitejohnson @seabbs

kaitejohnson:
This closes #107 and I believe #106. 

What it does:
- adds an additional argument to `wwinference()` that takes in an R seed for the initialization. This defaults to not having a seed
- removes the default seed of `123` in the `get_mcmc_options()`.
- adds a seed for reproducibility in the vignette
- documents this in the function documentation and in the vignette
- bug fix: `data-raw/test_data.R` didn't get updated to reflect expectation of log scale concentrations


vignettes/wwinference.Rmd
```diff
@@ -286,8 +286,12 @@ settings (see the documentation for `get_mcmc_options()`),
 however this vignette will use
 the default parameter settings which includes running 4 parallel chains with
 750 warm up iterations, 500 sampling iterations for each chain, a target average
-acceptance probability of 0.95 and a maximum tree depth of 12. The user may wish
-to adjust these as they are iterating to reduce model run-time or to achieve
+acceptance probability of 0.95 and a maximum tree depth of 12. The default is
+not to set a the seed for the random number generator for the MCMC model runs
+(which would produce stochastic results each time the model is run), but for
+reproducibility we will set the seed of the Stan PRNG to `123`in this vignette.
```
dylanhmorris, line 292:
```suggestion
reproducibility we will set the seed of the Stan PRNG to `123` in this vignette.
```

R/wwinference.R
```diff
@@ -189,6 +196,7 @@ wwinference <- function(ww_data,
   )
 
   init_lists <- NULL
+  set.seed(initial_values_seed)
```
dylanhmorris, line 199: Use `withr::with_seed` here instead of `set.seed`. See https://withr.r-lib.org/articles/withr.html#its-dangerous-to-change-state
```suggestion
  set.seed(initial_values_seed)
```

dylanhmorris: specifically, change this:

```R
    init_lists <- c()
    for (i in 1:fit_opts$n_chains) {
      init_lists[[i]] <- get_inits_for_one_chain(stan_data_list, params)
    }
```

to this

```R
withr::with_seed(initial_values_seed, {
     init_lists <- lapply(1:fit_opts$n_chains, \(x){get_inits_for_one_chain(stan_data_list, params)}
   })

**Review by dylanhmorris — CHANGES_REQUESTED** [review]
A few tweaks needed. Most importantly, use `withr::with_seed` instead of `set.seed`

File: vignettes/wwinference.Rmd [hunk 3]
```diff
@@ -307,8 +307,12 @@ settings (see the documentation for `get_mcmc_options()`),
...
+The user may wish to to adjust these settings as they are iterating to reduce
+model run-time or to achieve
 better convergence on a real-world use case.
```
dylanhmorris, line 316:
```suggestion

When applying the model to real data, experimenting with these MCMC settings may make it possible
to achieve improved model convergence and/or faster model fitting times. See the [Stan User's Guide](https://mc-stan.org/docs/cmdstan-guide/diagnose_utility.html#building-the-diagnose-command) for an introduction to No-U-Turn sampler convergence diagnostics and configuration parameters.   
```

R/wwinference.R
kaitejohnson: @dylanhmorris I am not following why we would use the :
```r
if(is.null(initial_values_seed)) {
   initial_values_seed <- sample.int(.Machine$integer.max, 1L)
}
```

Isn't the whole point that you want to be able to run :
```r
withr::with_seed(5, {
    fit <- do.call(
      wwinference::wwinference,
      model_test_data
    )
  })
```
And have that seed be passed silently to R, so we'd really want something like:
```r
 if (generate_initial_values) {
     if(is.null(initial_values_seed){

      init_lists <- lapply(
          1:fit_opts$n_chains,
             \(x) {
              get_inits_for_one_chain(stan_data_list)
            }
        )
    }else{
       withr::with_seed(initial_values_seed, {
        init_lists <- lapply(
           1:fit_opts$n_chains,
           \(x) {
             get_inits_for_one_chain(stan_data_list)
          }
        )
      })
   }
```

kaitejohnson: Update: 
 f2f discussion explaining that this:
 ` initial_values_seed <- sample.int(.Machine$integer.max, 1L)` would either be random if no seed set outside of the function or produce the same `initial_values_seed` reproducibly if the seed was set outside of the function, achieving the goal of allowing for both functionalities and stochastic by default

Issue #98, opened 2024-08-26
 Fix `check_req_ww_cols_present`


kaitejohnson:
## Problem
The function did not work as expected. The error message was non-intuitive when 'site' was missing. It failed BEFORE it got to the checker message.

## Requirements
- [ ] fix `check_req_ww_cols_present`
- [ ] add a test to replicate and see what happens when `site` was missing

## Context
- user feedback when running with own data

dylanhmorris:
I think at least one issue here is that regex pattern (which I wrote; my apologies) for `autoescape_brackets()` isn't quite right.

```R
autoescape_brackets <- function (string)
{
    return(gsub("\\{(.*)\\}", "{{\\1}}", string))
}
```
The clue is in the error message that @kaitejohnson got. Note the trailing bracket.
```R
Names must include the elements {'log_genome_copies_per_ml','lod','date','site','lab','site_pop'}, but is
missing elements site}
```

The problem is that `checkmate::check_names()` messages are of the form:

```R
Names must include the elements {'a', 'b', 'c'}, but is
missing elements {'a', 'b'}
```

```R
gsub("\\{(.*)\\}", "{{\\1}}", "{a}, {b}")
```
yields: 

```R
[1] "{{a}, {b}}"
```

This is because as written the pattern matches and replaces the _outermost_ brackets, not pairs of brackets reading left to right.

On reflection, I'm not sure the brackets are needed in the error message. We could just get rid of them (`gsub("\\{|\\}", "", string)`:

```R
Names must include the elements 'log_genome_copies_per_ml','lod','date','site','lab','site_pop', but is
missing elements 'site'
```

But if we do want the brackets, the simplest solution is probably sequential `gsub`: `gsub(,"\\}", "}}, gsub("\\{", "{{", string))`.

kaitejohnson:
I think I'm in favor of getting rid of all of the brackets! I will rewrite the `autoescape_brackets()` function! Thanks for the debugging @dylanhmorris I missed that checkmates returns two things of brackets!

gvegayon:
I think this would also work:

```r
> gsub("\\{([^}]+,?)\\}", "{{\\1}}", "{a}, {b}") 
```

Yields

```r
[1] "{{a}}, {{b}}"
```

And

```r
gsub("\\{([^}]+,?)\\}", "{{\\1}}", "{a,b}, {b}") 
```

yields

```r
[1] "{{a,b}}, {{b}}"
```

[...]

PR #127, opened 2024-08-31
Fix check for required wastewater columns

kaitejohnson:
This is aimed at addressing #98. It does the following:

- renames `check_req_ww_cols_present` to `assert_req_ww_cols_present` for consistency with `checkmate()` which uses assert to throw an error if condition isn't met (whereas check will return TRUE if met and a message (but not an errror) if not met).
- adds a test to our testing suite to exclude `site` from the testing data. This returns the desired error message: `Error:
! Required columns are missing from the wastewater data.
Names must include the elements {'log_genome_copies_per_ml','lod','date','site','lab','site_pop'}, but is
missing elements site}`

Unfortunately, I wasn't able to recreate the problem originally described by @akeyel 

Here's a reprex of what currently happens if I exclude site:
```r
library(wwinference)
 x <- tibble::tibble(
      date = lubridate::ymd("2024-01-01"),
      lab = 2,
      log_genome_copies_per_ml = 4,
      lod = 6,
      site_pop = 100,
      loc = "example"
    )
    conc_col_name <- "log_genome_copies_per_ml"
    lod_col_name <- "lod"
     assert_req_ww_cols_present(
        x,
        conc_col_name,
        lod_col_name
      )

Error:
! Required columns are missing from the wastewater data.
Names must include the elements {'log_genome_copies_per_ml','lod','date','site','lab','site_pop'}, but is
missing elements site}
```

@akeyel Would you be able to share a reprex to generate the error you were seeing when using this function?


akeyel:
@kaitejohnson, your reprex runs and provides the expected error on my computer as well.

I'm not sure what is causing the unexpected error, so haven't figured out how to make a reprex. Below is an example error I get, which goes away when the site and lab columns are added.

I tried your reprex with a data frame instead of a tibble as an input, but that was not the problem. I tried it with additional junk columns with NA values, but that also was not the problem. I tried adding the column names from my data set to your reprex, with NA values, but that also did not generate the error. I also tried adding a second row to the reprex, to see if it was a dimensionality thing, but that ran as expected also. So I'm not sure what in the larger data set is triggering it. 

image

kaitejohnson:
Thanks for sharing this! This is super helpful @akeyel 

I suspect that the error is coming from the way that we are pasting together the error messaging, though I am very confused as to why it would differ from one dataset to another... 

We use the `[checkmate](https://mllg.github.io/checkmate/#:~:text=Enter%20checkmate.%20Virtually%20every%20standard%20type%20of%20user%20error%20when)` package to check that required column names that are created as a vector of strings are in the dataset. This either returns TRUE, or an error message with `{}` that indicates the columns that are missing. Which is a really helpful error message!

However, when passing that to `cli::cli_abort()` it uses `glue` under the hood, and glue treats brackets as special characters (it expects them to be replaced by variables). So we wrote this utility function, `autoescape_brackets()` that replaces the `{` with `{{` which I believe should allow it to be rendered properly... 

To see if this is the problem, I am going to try removing the checkmate error message in this PR. Would you mind reinstalling the package on this branch and then testing your dataset? Or you could send over a fake dataset and I can test myself! Not sure why the reprex isn't producing it!

kaitejohnson:
Ok after seeing @dylanhmorris's suggestion in #98 I think the fix is probably just to get rid of the brackets.

I rewrote the `autoescape_brackets()` function to replace brackets with an empty string. When I take the `ww_data` and remove the  site column:

```r
ww_data <- wwinference::ww_data
ww_data_test <- ww_data |>dplyr::select(-site)
assert_req_ww_cols_present(ww_data_test, conc_col_name = "log_genome_copies_per_ml", lod_col_name = "log_lod")
Error: ! Required columns are missing from the wastewater data.
Names must include the elements 'log_genome_copies_per_ml','log_lod','date','site','lab','site_pop', but is missing elements 'site'
```

dylanhmorris (approved):
LGTM, thanks @kaitejohnson!

PR #134, opened 2024-09-01
Tweaks to model definition

### LINKED ISSUE #135: Tweaks to model definition
## Goal
A few suggestions brought up by @ChiragKumar9 
- [x] center and put on own line the renewal equation
- [x] infection feedback should sum up to $T_g$

kaitejohnson:
This PR closes #135

### DIFF
--- a/model_definition.md
+++ b/model_definition.md
@@ -22,7 +22,9 @@ See the [notation](#appendix-notation) section for an overview of the mathematic
 #### Renewal process for incident infections
 
 This component assumes that latent (unobserved) _expected_ incident infections per capita $I(t)$ are generated from a renewal [^Cori][^EpiNow2][^Epidemia] process described by:
+
 $$I(t) = \mathcal{R}(t) \sum_{\tau = 1}^{T_g} I(t-\tau) g(\tau)$$
+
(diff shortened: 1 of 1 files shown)


ChiragKumar9 (dismissed):
Beautiful

model_definition.md
```diff
@@ -67,8 +69,6 @@ The total number of subpopulations is then $K_\mathrm{total} = K_\mathrm{sites}
 This amounts to modeling the wastewater catchments populations as approximately non-overlapping; every infected individual either does not contribute to measured wastewater or contributes principally to one wastewater catchment.
 This approximation is reasonable if we restrict our analyses to primary wastewaster treatment plants, which avoids the possibility that an individual might be sampled once in a sample taken upstream and then sampled again in a more aggregated sample taken further downstream.
 
-If the sum of the wastewater site catchment populations meets or exceeds the reported jurisdiction population ($\sum\nolimits_{k=1}^{K_\mathrm{sites}} n_k \ge n$) we do not use a final subpopulation without sampled wastewater. In that case, the total number of subpopulations $K_\mathrm{total} = K_\mathrm{sites}$.
```
dylanhmorris, line 70: Why was this removed?

kaitejohnson: I wrote this out in a new issue #136  but am realizing now that we should leave this in and address as a separate PR, because we do currently support the functionality of having `sum(ww_pop)> total_pop`. 

I think we should warn the user loudly about this, as while it is the case for one jurisdiction in the US, I think in most cases it is likely an error in how the populations are being assigned...

PR #146, opened 2024-09-05
Set seeds in test_get_stan_data

### LINKED ISSUE #142: Fix `test_get_stan_data` so random failure doesn't occur
## Problem

Every so often this test fails despite setting the seed for reproducibility. I believe is because we are using `rnorm()` to generate concentration values and then checking for outliers, and then checking for the number of rows in the final dataset. Have seen it a few times in CI. My inclination is that the test is hard coded and should be made more robust.

dylanhmorris:
This should close #142

### DIFF
--- a/tests/testthat/test_get_stan_data.R
+++ b/tests/testthat/test_get_stan_data.R
@@ -1,17 +1,17 @@
-seed <- 123
-
-ww_data <- tibble::tibble(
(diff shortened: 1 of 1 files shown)


kaitejohnson (approved):
LGTM, thanks for the catch!

PR #153, opened 2024-09-09
Adding new class and method for get_draws

gvegayon:
- [x] The output from `get_draws_df()` will now be a list of class `wwinference_fit_draws` (S3 class).
- [x] The function `get_draws_df()` will be replaced by `get_draws()`.
- [x] The function will have an optional argument called `what` that will receive the strings `"all"`, and whatever other elements can be extracted via `tidy_bayes::spread_draws()`.
- [x] (optional) `get_draws_df()` will still work but will be an alias to `get_draws()` using the `.Deprecated()` function call; this will throw a warning message to the users indicating the function `get_draws_df()` will be removed in the next version of the model.
- [x] The proposed methods for the new class will be `plot()` and `print()`, with the `plot()` method having a `what=` argument that will specify which bits of the list to plot using the existing plotting functions.
- ~~[ ] `wwinference_fit_draws` will also have a method `as.data.frame()` to turn the list into a data frame.~~
- [x] A reference to this new feature will be added to the vignette.
- [x] A test will be added to ensure the function performs as expected when calling `what="all"`.


gvegayon:
This is ready for review, @kaitejohnson and @dylanhmorris. I changed my mind regarding the `as.data.frame` method as I think (i) it is overkill and (ii) I doubt about its usefulness. LMK if you think we should still have that.

R/sysdata.rda
gvegayon, line 1: What is this, and why do we need it, @kaitejohnson?

dylanhmorris: https://r-pkgs.org/data.html#sec-data-sysdata

dylanhmorris: In the case of this package, I believe it is ithe internal testing data https://github.com/CDCgov/ww-inference-model/blob/main/data-raw/test_data.R

kaitejohnson: Yes it is the internal testing data that is generated from running `data-raw/test_data.R`. We should probably have developers documentation on this workflow somewhere...

R/get_draws.R
```diff
@@ -0,0 +1,399 @@
...
+                                 ...,
+                                 what = "all") {
+  # Checking we are getting all
+  what_ok <- get_draws_what_ok
+
+  if (any(!what %in% what_ok)) {
+    idx <- which(!what %in% what_ok)
+    stop(
+      "The following invalid values were passed to `what`: ",
+      paste(what[idx], collapse = ", "), ". Valid values incllude: ",
```
kaitejohnson, line 96:
```suggestion
      paste(what[idx], collapse = ", "), ". Valid values include: ",
```

R/get_draws.R
```diff
@@ -0,0 +1,399 @@
...
+  stop(
+    "No method defined for get_draws for object of class(es) ",
+    paste(class(x), collapse = ", "),
+    ". Use directly on a wwinference_fit object or a",
+    "dataframe of wastewater observations.",
+    call. = FALSE
+  )
+}
+
+get_draws_what_ok <- c(
```
kaitejohnson, line 71: Is it standard practice to just define a vector like this outside of a function in one of these files? Will defer to you but I haven't seen this before!

R/get_draws.R
```diff
@@ -0,0 +1,399 @@
...
+  "all", "predicted_counts", "predicted_ww", "global_rt", "site_level_rt"
```
kaitejohnson, line 72: So my take here is that we made the decision to expect the user to specify snake case variable names, that do not necessarily perfectly align with the variable name in stan.  I am fine with this. 

I think `site_level_rt` should be `subpop_rt` (even though it is called `r_site_t` in the stan model).

R/get_draws.R
```diff
@@ -0,0 +1,399 @@
...
+#' @param global_rt Global Rt()
+#' @param site_level_r Site-level Rt()s
+#' @noRd
+new_wwinference_fit_draws <- function(
+    predicted_counts,
+    predicted_ww,
+    global_rt,
+    site_level_rt) {
+  # Checking colnames: Must match all exactly
+  predicted_counts_cnames <- c(
```
kaitejohnson, line 297: Pedantic but can we call these `colnames` not `cnames`?

R/get_draws.R
```diff
@@ -0,0 +1,399 @@
...
+#' @rdname get_draws
+#' @param x An object of class `get_draws`.
+#' @param y Ignored in the the case of `plot`.
+#' @details
+#' The plot method for `wwinference_fit_draws` is a wrapper of
+#' `get_plot_forecasted_counts`, `get_plot_ww_conc`, `get_plot_global_rt`,
+#' and `get_plot_subpop_rt`. Depending on the value of `what`, the function
+#' will call the appropriate method.
+#'
+plot.wwinference_fit_draws <- function(x, y = NULL, what, ...) {
```
kaitejohnson, line 362: What happens here if `what == "all"` ? or `what = c("predicted_counts", "predicted_ww")`?

R/get_draws.R
```diff
@@ -0,0 +1,399 @@
...
+#' `x` and `count_data`
+#' @export
+get_draws.data.frame <- function(x,
+                                 count_data,
+                                 stan_data_list,
+                                 fit_obj,
+                                 ...,
+                                 what = "all") {
+  # Checking we are getting all
+  what_ok <- get_draws_what_ok
```
kaitejohnson, line 90: How is this a check?

R/get_draws.R
```diff
@@ -0,0 +1,399 @@
...
+
+  if (any(!what %in% what_ok)) {
+    idx <- which(!what %in% what_ok)
+    stop(
+      "The following invalid values were passed to `what`: ",
+      paste(what[idx], collapse = ", "), ". Valid values incllude: ",
+      paste(what_ok, collapse = ", "), "."
+    )
+  }
+
```
kaitejohnson, line 100: Can you describe this logic a bit more? 
My thought is that we would just want, before the creation of each df, an if statement that checks for whether that
output variable is in the vector of `what`
```
if("predicted_counts" %in% what || what == "all") {
     make data frame
  }else{
  NULL
}
```

kaitejohnson: And then an `rlang::argmatch()` or some other test to ensure that `what` only contains either `all` or the output types we want

kaitejohnson (changes requested):
@gvegayon Looks really good, just a few comments for clarity and a pedantic request about naming conventions. I am not fully following the logic on the `what[{var_name}]` and proposed a solution that I find slightly easier to read. But happy to be convinced otherwise.

R/get_draws.R
gvegayon: Addressed.

R/figures.R
```diff
@@ -30,10 +30,23 @@ get_plot_forecasted_counts <- function(draws,
...
+        "The number of draws to plot (%i) should be less or equal to ",
+        n_draws_to_plot
+      ),
+      sprintf(
+        "the number of draws in the data (%i).",
+        n_draws_available
+      )
+    )
+  }
+
```
kaitejohnson, line 46: Suggest adding a test to ensure the correct df was passed into the function e.g.:


```suggestion

data_type <- draws |> dplyr::distinct("name") |> dplyr::pull()
if(data_type != "predicted_counts"){
    cli::cli_abort("Data passed to plotting function does not contain the expected variable type, `predicted_counts`")
}
```suggestion

Comment (kaitejohnson) [reply R/figures.R:46]: same for the other plotting functions

File: R/get_draws.R [hunk 9]
```diff
@@ -0,0 +1,444 @@
...
+    idx <- which(!what %in% what_ok)
+    stop(
+      "The following invalid values were passed to `what`: ",
+      paste(what[idx], collapse = ", "), ". Valid values include: ",
+      paste(what_ok, collapse = ", "), "."
+    )
+  }
+
+  what_ok <- logical(length(what_ok))
+  names(what_ok) <- get_draws_what_ok
```
kaitejohnson, line 104: I realized when working on #158 that if `stan_data_list$include_ww = 0` than we should not be allowing the user to get `subpop_rt` and `pred_ww`. The model as written currently will still return the prior, but I refactored it so that we are not carrying around an unnecessarily large model/latent variables for the hospital admissions only case, which means expected subpopulation R(t) estimates and predicted wastewater will not by default be generated.  Even if we do change this, I don't think we want "all" to return all 4 if its not fit to wastewater data. 

We probably need to add another layer of logic then that says something like:
```suggestion
  names(what_ok) <- ifelse(stan_data_list$include_ww ==0,  c("all", "predicted_counts", "global_rt"), get_draws_what_ok)
```

R/get_draws.R
```diff
@@ -0,0 +1,444 @@
...
+#'
+plot.wwinference_fit_draws <- function(x, y = NULL, what, ...) {
+  if (length(what) != 1L) {
+    stop(
+      "The value provided to `what` must be a length one character vector. ",
+      "Currently, it is of length ", length(what), "."
+    )
+  }
+
+  which_what_are_ok <- setdiff(get_draws_what_ok, "all")
```
kaitejohnson, line 408: Again, probably want some logic here for the `stan_data_list$include_ww` to tell it to only plot the two outputs we want

kaitejohnson (changes requested):
@gvegayon This looks really good! I just have two minor comments, I think they could be separate PRs because they technically don't break things here. 

1. Is just a check in the figures for the correct name of the variable
2. Is to adjust what `get_draws()` prints/plots/outputs under the `include_ww = 0` case, so that we are not falsely implying a wastewater model was fit with subpopulation R(t)s. Right now these would come from the prior, later I think we will remove them see #158.

R/figures.R
gvegayon: I was adding this bit but realized this is not going to work as we don't have the column "name" anymore.

R/get_draws.R
kaitejohnson: Will open a separate issue for these two items!

kaitejohnson (approved):
Added two additional issues to address separately. This looks great thanks @gvegayon!

Issue #149, opened 2024-09-09
Refactor hierarchical site-level estimation to estimate `n_subpops-1` deviations from the "reference" subpopulation


kaitejohnson:
Edited to reflect additional intercept term and propose optional patchwork:

## Goal
This should address #136, but also a more fundamental issue brought up by @damonbayer and @sbidari, which is that the current version of the "hospital admissions only" model is still going to estimate `n_subpops` infection dynamics, even though this will not be informed by wastewater concentration data and should not have a patchwork or hierarchical population structure. Instead, the desired behavior here is that:
- the "hospital admissions only" model reverts to estimating the R(t) of a single reference population (here the total population). This will then approximate the `renewal_ww_hosp.stan` model in the [production/eval](https://github.com/CDCgov/wastewater-informed-covid-forecasting) repo. Here in effect there will be $n_{\mathrm{sites}} = 0$, and the population of the "auxiliary site of those not captured by wastewater" is the total population, $n$ 
- when $\sum\nolimits_{k=1}^{K_\mathrm{sites}} n_k \ge n$ we want to estimate `n_subpops = n_sites` R(t) estimates, whose per capita infections generate expected counts in the total pop. Here there will be no "auxiliary site", and the "reference" subpopulation will just be the largest wastewater catchment area. 
- when $\sum\nolimits_{k=1}^{K_\mathrm{sites}} n_k < n$ (the standard case) we will estimate the site level R(t) estimates as deviations from the reference subpopulation, which is by default the auxiliary site (the subpopulation not captured by wastewater). 

This results in the following changes to the model definition:

Currently, we estimate a global undamped effective reproductive number $\mathcal{R}^\mathrm{u}(t)$. We now instead will estimate a reference effective reproductive number $\mathcal{R}^0(t)$ with $K_{\mathrm{subpops}}-1$ deviations from the reference. 

The number of subpopulations falls under a few distinct cases:
-  there are $n_{\mathrm{sites}}$,  $\sum\nolimits_{k=1}^{K_\mathrm{sites}} n_k < n$, therefore $K_{\mathrm{subpops}} = n_{\mathrm{sites}}+ 1$.  If $n_{\mathrm{sites}} = 0$, i.e. there is no wastewater data, then there is only $K_{\mathrm{subpops}} =1$ and it is the reference subpopulation.  The reference is the subpopulation not covered by wastewater by default
[...]

Issue #145, opened 2024-09-05
Discuss future of MCMC snapshot test in `test_ww_model.R`


dylanhmorris:
Will add some details here.

athowes:
What are the details here? (Personally have not used snapshot tests yet but interested as to their utility.)

PR #158, opened 2024-09-10
Restructure hierarchical estimation based on reference subpopulation

kaitejohnson:
This is a WIP. Tasks include:
- [X]  refactor stan model to estimate `n_subpop` infection dynamics, with the first `subpop` being the reference subpop
- [X] add a new parameter $m$, which represents the shift of the reference population from the central dyamic
- [X] This also includes getting rid of language that interchangeably uses sites as subpops. Observations occur in lab-site pairs
- [X] refactor `get_stan_data()` to appropriately index the observations 
- [X] refactor `ww_inference`  to specify that when `include_ww=0`, `ww_data` should be empty
- [X] add tests for the 3 use cases described in #149 that ensure the stan data is coded properly
- [X] added logic to `get_draws()` so that if `include_ww == 0`, `predicted_ww` and `subpop_rt` are not returned if the user passes `all`, and an error message is returned if they asked for either of those outputs.
- [X] update test data because of new model formulation 
- [X] revise model definition 
- [X] internal bookkeeping refactor (see below, creation of standardized mappings that we use to generate stan data and pass as an output for post processing)

Some scope creep:
- [X] parameter diagnostics and summary diagnostics seemed to have gotten swamped in their naming. Edited accordingly so `summary_diagnostics()` spits out the number of divergent transitions in each train, the number exceeding the max tree depth, and the EBFMI of each chain.  `parameter_diagnostics()` spits out the table of metadata on each model parameter (e.g. mean, median, stdev, ess, rhat)

As always anytime we change the model:
- [X] update test data
- [X] make sure tests pass


inst/stan/wwinference.stan
```diff
@@ -35,15 +35,14 @@ data {
...
   real<lower = state_pop> norm_pop;
   array[owt] int<lower=1,upper=ot + ht> ww_sampled_times; // a list of all of the days on which WW is sampled
-                                   // will be mapped to the corresponding sites (ww_sampled_sites)
+                                   // will be mapped to the corresponding subpops (ww_sampled_subpops)
   array[oht] int<lower=1, upper=ot> hosp_times; // the days on which hospital admissions are observed
-  array[owt] int<lower=1,upper=n_subpops> ww_sampled_sites; // vector of unique sites in order of the sampled times
-  array[owt] int<lower=1,upper=n_ww_lab_sites> ww_sampled_lab_sites; // vector of unique lab-site combos i
-   	     // n order of the sampled times
+  array[owt] int<lower=1,upper=n_subpops> ww_sampled_subpops; // vector of unique subpops in order of the sampled times
+  array[owt] int<lower=1,upper=n_ww_lab_sites+1> ww_sampled_lab_sites; // vector mapping the subpops to lab-site combos
```
dylanhmorris, line 41: Presumably the `n_ww_lab_sites+1`th index corresponds to an auxiliary site without observed wastewater. But given that this vector associates wastewater observations to lab_site combos, that index should never be used and should raise an error if passed in.

```suggestion
  array[owt] int<lower=1,upper=n_ww_lab_sites> ww_sampled_lab_sites; // vector mapping the subpops to lab-site combos
```

inst/stan/wwinference.stan
```diff
@@ -111,21 +111,23 @@ transformed data {
...
   real<lower=0, upper=1> autoreg_rt;// coefficient on AR process in R(t)
-  real log_r_mu_intercept; // state-level mean baseline reproduction number estimate (log) at t=0
+  real log_r_0_intercept; // reference subpop baseline reproduction number estimate (log) at t=0
   real<lower=0> sigma_rt; // magnitude of site level variation from state level
-  real<lower=0, upper=1> autoreg_rt_site;
+  real<lower=0, upper=1> autoreg_rt_subpop;
   real<lower=0, upper=1> autoreg_p_hosp;
-  matrix[n_subpops, n_weeks] error_site; // matrix of subpopulations
+  matrix[include_ww ? n_subpops-1 : 1, n_weeks] error_subpop;
+  //matrix[n_subpops-1, n_weeks] error_subpop; // matrix of subpopulation deviations from ref R(t)
```
dylanhmorris, line 123: Commented code
```suggestion
```

inst/stan/wwinference.stan
```diff
@@ -111,21 +111,23 @@ transformed data {
...
   real<lower=0,upper=1> i_first_obs_over_n; // per capita
-  // infection incidence on the day of the first observed infection
-  vector[n_subpops] eta_i_first_obs; // z-score on logit scale of site
+  // infection incidence on the day of the first observed infection of reference subpop
```
dylanhmorris, line 125:
```suggestion
  // mean per capita infection incidence on the day of the first observation
```
It's not an observed infection. Apologies for not catching this before.

inst/stan/wwinference.stan
```diff
@@ -111,21 +111,23 @@ transformed data {
...
+  vector[n_subpops-1] eta_i_first_obs; // z-score on logit scale of site
```
dylanhmorris, line 126:
```suggestion
  vector[n_subpops] eta_i_first_obs; // z-score on logit scale of subpopulation per capita infection incidences on the day of the first observation
```

inst/stan/wwinference.stan
```diff
@@ -111,21 +111,23 @@ transformed data {
...
   // initial per capita infection incidence relative to state value
   real<lower=0> sigma_i_first_obs; // stdev between logit state and site initial
   // per capita infection incidence
-  vector[n_subpops] eta_initial_exp_growth_rate; // z scores of individual site level initial exponential growth rates
+  vector[n_subpops-1] eta_initial_exp_growth_rate; // z scores of individual site level initial exponential growth rates
```
dylanhmorris, line 130:
```suggestion
  vector[n_subpops] eta_initial_exp_growth_rate; // z scores of individual subpopulation-level initial exponential growth rates
```

inst/stan/wwinference.stan
```diff
@@ -111,21 +111,23 @@ transformed data {
...
   real<lower=0> sigma_initial_exp_growth_rate; // sd of distribution of site level initial exp growth rates
```
dylanhmorris, line 131:
```suggestion
  real<lower=0> sigma_initial_exp_growth_rate; // sd of distribution of subpopulation level initial exp growth rates
```

inst/stan/wwinference.stan
```diff
@@ -111,21 +111,23 @@ transformed data {
...
   real mean_initial_exp_growth_rate; // mean of distribution of site level initial exp growth rates
```
dylanhmorris, line 132:
```suggestion
  real mean_initial_exp_growth_rate; // mean of distribution of subpopulation level initial exp growth rates
```

dylanhmorris (commented):
Before I review further, I think you want to keep things like `i_first_obs` and `initial_exp_growth_rate` as hierarchical about an inferred central value (so the reference subpopulation has z-score just like everyone else. It's only for $\mathcal{R}(t)$ that we want things to be relative to the reference subpop, rather than an inferred central value.

inst/stan/wwinference.stan
kaitejohnson: Oh yes sorry this was an artifact from the way I was setting it up initially, then changed things around

kaitejohnson:
> Before I review further, I think you want to keep things like `i_first_obs` and `initial_exp_growth_rate` as having an implicit central value that is not necessarily the same as the reference population's value. So you'll want offsets for the reference population for those too.
> 
> You'll also want to only infer and apply those offsets the number of subpopulations is > 1, otherwise they should be fixed at 0.

Good catch, I added the offset as the final step and forgot about the other site level parameters. Will edit.

kaitejohnson:
Updated, though I wasn't sure exactly what good priors on these offsets would be, in particular `m_first_obs_stdev` other than very small to ensure that the value samples for I0/N doesn't go above 1 or below 0...

kaitejohnson:
Test are passing locally from my Mac. Will try regenerating from WSL2 (was frozen..)

inst/stan/wwinference.stan
```diff
@@ -160,80 +169,112 @@ transformed parameters {
...
-				 eta_sd,
-				 w,
-				 0);
-  unadj_r = ind_m*log_r_mu_t_in_weeks;
+  // If only one subpopulation, then we don't want to estimate any offsets
+  if(n_subpops == 1){
+    m_tr = 0;
+    m_first_obs_tr = 0;
+    m_r_tr = 0;
+  }
```
dylanhmorris, line 193:
```suggestion
```

dylanhmorris (commented):
Started leaving some thoughts on how to structure the Stan code, but realized it might be clearer to make a branch off this branch and PR in, which I am doing now.

inst/stan/wwinference.stan
```diff
@@ -111,23 +113,26 @@ transformed data {
...
+  real m_first_obs; // shift of unbserved initial infections per capita of reference pop from central dynamic
   real<lower=0, upper=1> autoreg_rt;// coefficient on AR process in R(t)
-  real log_r_mu_intercept; // state-level mean baseline reproduction number estimate (log) at t=0
+  real log_r_0_intercept; // reference subpop baseline reproduction number estimate (log) at t=0
   real<lower=0> sigma_rt; // magnitude of site level variation from state level
-  real<lower=0, upper=1> autoreg_rt_site;
+  real<lower=0, upper=1> autoreg_rt_subpop;
   real<lower=0, upper=1> autoreg_p_hosp;
-  matrix[n_subpops, n_weeks] error_site; // matrix of subpopulations
+  matrix[include_ww ? n_subpops-1 : 1, n_weeks] error_subpop;
```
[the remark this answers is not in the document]
kaitejohnson: I think this has to be 1 because of the initialization (you can't initialize a [0,0] matrix).... it shouldn't get used.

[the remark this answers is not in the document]
dylanhmorris: In that case use an array of vectors?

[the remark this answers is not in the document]
kaitejohnson: Yeah I have never found a good solution to this other than setting it to not be 0 and not using it. It works ok for vectors but not for matrices because a [0,0] matrix in R gets passed to stan as a length 0 vector... recognize this as a hacky work around so happy to hear otherwise if you are able to get something to work for the initialization.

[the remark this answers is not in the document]
dylanhmorris: pass `NULL`?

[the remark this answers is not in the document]
kaitejohnson: I need to play around with options but I think stan still give you a warning if an initial value is missing even for a parameter specified as [0,0]

dylanhmorris:
#168 is now ready for review @kaitejohnson

gvegayon:
Current R CMD check error here:

```bash
Running the tests in ‘tests/testthat.R’ failed.
Last 13 lines of output:
              sd_log_sigma_ww_site_prior_sd = 0.693, t_peak_mean = 5L, 
              t_peak_sd = 1L, viral_peak_mean = 5.1, viral_peak_sd = 0.5, 
              ww_site_mod_sd_sd = 0.25)), fit_opts = list(iter_warmup = 25, 
          iter_sampling = 25, n_chains = 1, seed = 55, adapt_delta = 0.95, 
          max_treedepth = 12), generate_initial_values = TRUE)`: argument "ww_data" is missing, with no default
  Backtrace:
      ▆
   1. ├─withr::with_seed(...) at test_no_ww_model.R:9:3
   2. │ └─withr::with_preserve_seed(...)
   3. ├─base::do.call(wwinference::wwinference, model_test_data_no_ww) at test_no_ww_model.R:10:5
   4. └─wwinference (local) `<fn>`(...)
  
  [ FAIL 1 | WARN 3023 | SKIP 0 | PASS 242 ]
  Error: Test failures
  Execution halted
* checking for non-standard things in the check directory ... OK
* checking for detritus in the temp directory ... OK
* DONE

Status: 1 ERROR, 1 NOTE
```

kaitejohnson:
> Current R CMD check error here:
> 
> ```shell
> Running the tests in ‘tests/testthat.R’ failed.
> Last 13 lines of output:
>               sd_log_sigma_ww_site_prior_sd = 0.693, t_peak_mean = 5L, 
>               t_peak_sd = 1L, viral_peak_mean = 5.1, viral_peak_sd = 0.5, 
>               ww_site_mod_sd_sd = 0.25)), fit_opts = list(iter_warmup = 25, 
>           iter_sampling = 25, n_chains = 1, seed = 55, adapt_delta = 0.95, 
>           max_treedepth = 12), generate_initial_values = TRUE)`: argument "ww_data" is missing, with no default
>   Backtrace:
>       ▆
>    1. ├─withr::with_seed(...) at test_no_ww_model.R:9:3
>    2. │ └─withr::with_preserve_seed(...)
>    3. ├─base::do.call(wwinference::wwinference, model_test_data_no_ww) at test_no_ww_model.R:10:5
>    4. └─wwinference (local) `<fn>`(...)
>   
>   [ FAIL 1 | WARN 3023 | SKIP 0 | PASS 242 ]
>   Error: Test failures
>   Execution halted
> * checking for non-standard things in the check directory ... OK
> * checking for detritus in the temp directory ... OK
> * DONE
> 
> Status: 1 ERROR, 1 NOTE
> ```

Yeah I am going to expand this test a bit to include 1. passing ww but setting include_ww = 0 2. passing no ww and setting include_ww = 0 and 3. passing no ww and setting include_ww = 1. The model should be able to handle all three cases, but it can currently only handle 1!

kaitejohnson:
Relatedly, I would love to suppress these warnings (either adding more iterations or ideally just telling cmdstan to suppress them)... will look into it

dylanhmorris:
> Relatedly, I would love to suppress the ESS warnings (either adding more iterations or ideally just telling cmdstan to suppress them)... will look into it

Worth discussing in #145

kaitejohnson:
looks like ww tests are failing again, will try regenerating test data...

R/get_draws.R
```diff
@@ -169,6 +173,13 @@ get_draws.data.frame <- function(x,
 
 
   ww_draws <- if (what_ok["predicted_ww"]) {
+    # Lab-site index to corresponding lab, site, and site population size
+    lab_site_spine <- x |>
+      dplyr::distinct(
+        .data$site, .data$lab, .data$lab_site_index,
+        .data$site_pop, .data$lab_site_name
```
dylanhmorris, line 180:
```suggestion
        .data$site, 
        .data$lab, 
        .data$lab_site_index,
        .data$site_pop, 
        .data$lab_site_name
```

R/get_draws.R
```diff
@@ -228,43 +239,58 @@ get_draws.data.frame <- function(x,
...
+        dplyr::mutate(subpop_index = .data$subpop_index + 1) |>
+        dplyr::bind_rows(tibble::tibble(
+          site = NA,
+          subpop_index = 1,
+          subpop_pop = stan_data_list$subpop_size[1],
+          subpop = "remainder of pop"
+        ))
+    } else {
+      subpop_spine <- site_spine
+    }
```
dylanhmorris, line 274: Now that we're making a strong distinction between subpopulations and sites, I think it's clearer and safer to do this with an explicit `subpop_index` and `site_index` defined in pre-processing.  This also makes it possible to adjust whether auxiliary site == reference site if we wish.
 
Related: should we not just calculate the spines in pre-processing and have them as attributes of the `wwinferencefit` object?

dylanhmorris (changes requested):
Some suggestions and questions @kaitejohnson

R/get_stan_data.R
```diff
@@ -282,15 +267,6 @@ get_stan_data <- function(input_count_data,
...
-      "The sum of the wastewater site catchment area populations:",
-      "is greater than the global population. While the model supports this",
-      "we advise checking your input data to ensure it is specified correctly."
-    ))
-  }
-
-
   # Logic to determine the number of subpopulations to estimate R(t) for:
   # First determine if we need to add an additional subpopulation
   add_auxiliary_site <- ifelse(pop >= sum(ww_values$pop_ww), TRUE, FALSE)
```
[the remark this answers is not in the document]
kaitejohnson: Yes good point.

R/get_draws.R
kaitejohnson: Just to make sure I am understanding correctly, in this implementation you are suggesting that we have an explicit subpop_index which here would be 1:(site_index) + 1 if theres an aux site and equal to site index if not. 

We moved away from passing the spines as attributes of `wwinferencefit`, with the idea being that now that we are returning the input data, any of the spines can be calculated from the data (plus passing several spines seemed just a bit clunky...)

R/validate.R
```diff
@@ -209,12 +208,44 @@ validate_both_datasets <- function(input_count_data,
...
+    sum()
+  total_pop <- input_count_data |>
+    dplyr::distinct(.data$total_pop)
+  if (sum_site_pops > total_pop) {
+    cli::cli_warn(c(
+      "The sum of the populations in the wastewater catchment areas is",
+      "larger than the total population. While the model supports this",
+      "we advise checking your input data to ensure it is specified",
+      "correctly and to make sure that wastewater treatment plants are not",
+      "overlapping"
```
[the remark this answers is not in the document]
kaitejohnson: Yes can reword

R/get_stan_data.R
```diff
@@ -351,6 +329,16 @@ get_stan_data <- function(input_count_data,
...
 
+  # if adding auxiliary site, then we need to adjust site and lab indices
+  # such that first subpopulation is the auxiliary site.
+  if (add_auxiliary_site) {
+    subpop_to_samples_map <- ww_indices$ww_sampled_sites + 1
+    lab_site_to_subpop_map <- ww_indices$lab_site_to_site_map + 1
+  } else {
+    subpop_to_samples_map <- ww_indices$ww_sampled_sites
+    lab_site_to_subpop_map <- ww_indices$lab_site_to_site_map
+  }
```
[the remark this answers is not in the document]
kaitejohnson: Are you saying an additional `site_to_subpop_map` as well as the `labs_site_to_subpop_map`. I would prefer not to remove the `lab_site_to_subpop_map` because this is the unit that maps latent infections to observations, whereas `site_to_subpop` only gets you somewhere intermediate.... (you'd have to additionally index site_to_lab_site)

So the mapping would look something like:
`exp_obs_conc[lab_site[i], t] = conc[site_to_subpop[site_to_lab[i]], t]` vs what we have now which is:
`exp_obs_conc[lab_site[i], t] = conc[lab_site_to_subpop[i]]`

R/initialization.R
```diff
@@ -96,5 +103,19 @@ get_inits_for_one_chain <- function(stan_data, stdev = 0.01) {
...
+  if (stan_data$n_subpops > 1) {
+    init_list$error_subpop <- as.matrix(
+      matrix(
+        stats::rnorm((n_subpops - 1) * n_weeks,
+          mean = 0,
+          sd = stdev
+        ),
+        (n_subpops - 1),
+        n_weeks
+      )
```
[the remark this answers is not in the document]
kaitejohnson: Ok maybe you meant something different, but I can't do something like:
```r
test <- list(x = 1, y = ifelse(FALSE, 0, NULL))

Error in ans[npos] <- rep(no, length.out = len_[npos] :
replacement has length zero
In addition: Warning message: 
In rep(no, length.out = len) : 'x' is NULL so the result will be NULL
```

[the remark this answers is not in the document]
kaitejohnson: Which is why CI is failing now :(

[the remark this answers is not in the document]
kaitejohnson: Reverted back to original implementation, removing the unnecessary `as.matrix()`

R/get_stan_data.R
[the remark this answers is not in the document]
kaitejohnson: Ok see my update:
```r
 # This creates a dataframe that maps subpopulations to sample indices
  # by joining the site:subpop map to vector of site indices for each sample.
  # Then we just pull the subpop indices corresponding to each sample
  subpop_to_samples_map <- tibble::tibble(
    site_index = ww_indices$ww_sampled_sites
  ) |>
    dplyr::left_join(subpop_data$site_to_subpop_map, by = "site_index") |>
    pull(subpop_index)
  # This creates a dataframe that maps subpopulations to lab_site indices
  # by joining the site:subpop map to lab_site indices. Then we just pull the
  # subpop indices corresponding to the lab_sites
  lab_site_to_subpop_map <- tibble::tibble(
    site_index = ww_indices$lab_site_to_site_map
  ) |>
    dplyr::left_join(subpop_data$site_to_subpop_map, by = "site_index") |>
    pull(subpop_index)
```

What I am doing here is making a tibble in `get_subpop_data()` that looks like:
```
# A tibble: 5 × 2
  site_index subpop_index
       <int>        <int>
1         NA            1
2          1            2
3          2            3
4          3            4
5          4            5
```
Then I am using it to get the indices I want here by joining this to the wastewater data by site index and passing the vector of correspinding subpop indices.

My proposal is that since this mapping + the `site_name` is needed in post processing + a mapping of dates to times and a mapping of subpops to lab_sites, I think these 3 mappings should be added to the output of wwinference. This way, nothing in postprocessing needs to be redone if we change how we do subpopulations or if there are multiple subpopulations without site level observations etc. 

The 3 mappings would be:
```
site_to_subpop_map <- 
site_index | subpop_index | subpop_name | site_name 

lab_site_to_subpop_map <-
lab_site_index | subpop_index | subpop_name | lab_site_name

date_to_t_map <-
t | date

**Comment by dylanhmorris** [comment]
@kaitejohnson still a few failing tests. I made one easy fix myself. Could you fix the others and re-request review?

**Comment by kaitejohnson** [comment]
Pretty significant refactor of internal workings so that we are passing around the following mappings:
date-time mapping:
```r
date_time_spine
# A tibble: 126 × 2
   date           t
   <date>     <int>
 1 2023-09-01     1
 2 2023-09-02     2
 3 2023-09-03     3
 4 2023-09-04     4
 5 2023-09-05     5
 6 2023-09-06     6
 7 2023-09-07     7
 8 2023-09-08     8
 9 2023-09-09     9
10 2023-09-10    10
# ℹ 116 more rows
# ℹ Use `print(n = ...)` to see more rows
```
lab-site subpop spine:
```r
A tibble: 5 × 9
  lab_site_index site_index  site   lab site_pop lab_site_name   subpop_pop subpop_index subpop_name
           <int>      <int> <dbl> <dbl>    <dbl> <glue>               <dbl>        <int> <chr>      
1              1          1     2     1   400000 Site: 2, Lab: 1     400000            2 Site: 2    
2              2          1     2     2   400000 Site: 2, Lab: 2     400000            2 Site: 2    
3              3          2     3     3   200000 Site: 3, Lab: 3     200000            3 Site: 3    
4              4          3     4     3   100000 Site: 4, Lab: 3     100000            4 Site: 4    
5              5          4     5     3    50000 Site: 5, Lab: 3      50000            5 Site: 5    
```
site subpop spine:
```r
site_subpop_spine
# A tibble: 5 × 5
  site_index  site subpop_pop subpop_index subpop_name            
       <int> <dbl>      <dbl>        <int> <chr>                  
1         NA    NA    2250000            1 remainder of population
2          1     2     400000            2 Site: 2                
3          2     3     200000            3 Site: 3                
4          3     4     100000            4 Site: 4                
5          4     5      50000            5 Site: 5    
```

The logic for creating the auxiliary subpopulation or setting the first and only subpopulation to the entire population if hospital admissions only is now only happening once in the `get_site_subop_spine()` function which gets called inside `wwinference()`. 

These get returned as `raw_input_data` in the `wwinference_fit` object and are used directly for post processing.

kaitejohnson:
I believe this is still failing at the `test_ww_model` test, despite regenerating test data. This has been a consistent issue, and tbh I am not really sure this test is providing us much value. Also very frustrating that the test data can't be created on a mac (running the tests locally froze my Ubuntu). 

Perhaps is a separate issue but just flagging.

R/get_draws.R
```diff
@@ -292,16 +285,20 @@ print.wwinference_fit_draws <- function(x, ...) {
 
   timepoints <- c(
```
dylanhmorris, line 286: This might be a good place for a comment clarifying the aim of this code

kaitejohnson:
Pending CI passes, this is ready for review.

R/get_stan_data.R
```diff
@@ -42,68 +42,230 @@ get_input_ww_data_for_stan <- function(preprocessed_ww_data,
...
+      ) |>
+      dplyr::arrange(.data$lab_site_index) |>
+      dplyr::distinct() |>
+      dplyr::mutate(
+        "lab_site_name" = glue::glue(
+          "Site: {site}, Lab: {lab}"
+        )
+      )
+  } else {
+    lab_site_site_spine <- tibble::tibble()
```
dylanhmorris, line 150: Do you want this to be an empty tibble or a zero-length tibble with the appropriate column names?

R/get_stan_data.R
```diff
@@ -42,68 +42,230 @@ get_input_ww_data_for_stan <- function(preprocessed_ww_data,
...
+        "site_index" = NA,
+        "site" = NA,
+        "site_pop" = total_pop - sum(site_indices$site_pop)
+      )
+    } else {
+      aux_subpop <- tibble::tibble()
+    }
+
+    site_subpop_spine <- aux_subpop |>
+      dplyr::bind_rows(site_indices) |>
```
dylanhmorris, line 201: iirc we discussed using the largest subpopulation (which will often be the auxiliary) as the reference by default, but also letting the user configure the reference population. As it stands, it will either be the auxiliary or the first site (whenever there is no auxiliary).

R/get_stan_data.R
```diff
@@ -228,14 +423,21 @@ get_stan_data <- function(input_count_data,
...
+      input_count_data = input_count_data,
+      input_ww_data = input_ww_data,
+      date_time_spine = date_time_spine,
+      lab_site_site_spine = lab_site_site_spine,
+      site_subpop_spine = site_subpop_spine,
+      lab_site_subpop_spine = lab_site_subpop_spine,
+      calibration_time = calibration_time,
+      forecast_date = forecast_date
+    )
+  }
```
dylanhmorris, line 438: is there count data validation we can do when `include_ww == 0`?

R/wwinference.R
```diff
@@ -159,12 +159,30 @@ wwinference <- function(ww_data,
...
+  if (is.null(ww_data) || nrow(ww_data) == 0) {
+    cli::cli_warn(
+      c(
+        "No wastewater data was passed to the model.",
+        "The model will default to fitting only to the count data"
+      )
+    )
+    include_ww <- 0
+  }
+  # If include_ww <-0, we will specify an empty dataset
```
dylanhmorris, line 180:
```suggestion
  # If include_ww == 0, we will specify an empty dataset
```

dylanhmorris (commented):
This looks very good, @kaitejohnson. A few minor questions.

R/get_stan_data.R
kaitejohnson: I think I want it to be empty, since there is no mapping in this case. I don't think there is a need for it to have the correct column names.

R/get_stan_data.R
kaitejohnson: Ok yeah, I think this configurability should be a separate issue. I think its out of scope for this PR and I'd have to think a bit about how we'd want to handle this on the user interface side.

R/get_stan_data.R
kaitejohnson: It already happens in the pre-processing of the count data

kaitejohnson: This is just an additional one that is only about how they interact with each other

R/get_stan_data.R
dylanhmorris: Sounds good. Create an issue?

dylanhmorris (approved):
LGTM. Great work, @kaitejohnson!

R/get_stan_data.R
kaitejohnson: https://github.com/CDCgov/ww-inference-model/issues/196

R/get_stan_data.R
kaitejohnson: https://github.com/CDCgov/ww-inference-model/issues/197

PR #164, opened 2024-09-13
163 expand R version 

kaitejohnson:
Attempting to run install of package with multiple R versions in CI. Not sure if there's a more elegant way to do this, but this seemed to work. I think it would make sense to always test both the minimum and latest R release?

### DIFF
--- a/.github/workflows/r-cmd-check.yaml
+++ b/.github/workflows/r-cmd-check.yaml
@@ -6,37 +6,14 @@ on:
     branches: [main, prod]
 
 jobs:
-  check-package-min-R-version:
+  check-package:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
     - uses: actions/checkout@v4
     - uses: r-lib/actions/setup-r@v2
       with:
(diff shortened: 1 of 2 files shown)


dylanhmorris (commented):
This will work and looks solid, but to DRYify it you can use this: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow

If you're inspired, give it a try. If not, we can merge as is and make an issue to replace this with that.

kaitejohnson:
Hmm I wonder if this only works on an array of numerics, whereas `r-version` expects only a string, number, or boolean (not an array):
image.

kaitejohnson:
I think this is ready for review but we'd want to change the github actions rulesets to `check-package` to `check-package (4.1.0)` and `check-package (release)` once merged

dylanhmorris (approved):
LGTM, thanks @kaitejohnson. I have updated the branch protection rule.

kaitejohnson:
TY! Going to do the same on the other repo

PR #165, opened 2024-09-16
Solution to L_Omega initialization 

kaitejohnson:
This isn't ideal, but I think is better than having warnings about no inits for L Omega if the correlation structure isn't the random correlation matrix.


cbernalz (approved):
This looks good, do you want me to merge it or you?

kaitejohnson:
You can merge @cbernalz !

PR #167, opened 2024-09-17
Update _pkgdown.yml

kaitejohnson:
Try using `math_render: katex`


dylanhmorris:
Can you do this into `main` instead? That way:

1. we have this fix everywhere
2. the workflow will actually run and we can look at the rendered results before merging

PR #169, opened 2024-09-17
fix rendering to katex, add mathcal Rt to vignette

kaitejohnson:
Sets the math rendering to `katex`, per solution described here https://github.com/r-lib/pkgdown/issues/1966#issuecomment-2132233271 (thanks @damonbayer for the link)

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -1,3 +1,4 @@
 url: https://cdcgov.github.io/ww-inference-model/
 template:
   bootstrap: 5
+  math-rendering: katex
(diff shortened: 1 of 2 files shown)


dylanhmorris (approved):
LGTM

kaitejohnson:
@dylanhmorris Whoops, this actually did not work on the `pkgdown` site, which is weird bc the vignette renders in Rstudio using pkgdown! https://cdcgov.github.io/ww-inference-model/articles/wwinference.html#fitting-the-model

PR #170, opened 2024-09-17
Tweaks to main vignette

kaitejohnson:
Fixes rendering to specify mathjax. Worked locally, but so did `katex`. Otherwise, will revert back to using $R(t)$ for now. 

Also fixing plot sizing.

### DIFF
--- a/R/figures.R
+++ b/R/figures.R
@@ -123,11 +123,11 @@ get_plot_ww_conc <- function(draws,
         color = .data$subpop,
         group = .data$draw
       ),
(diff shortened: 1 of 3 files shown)


dylanhmorris (approved):
LGTM. 

- Thanks for fixing the plots! They look a lot better.
- Math looks less nice than the `katex` version, so (low priority) we may want to explore trying to get that to work, but for now let's go with this as a fix.

kaitejohnson:
Ooh yeah I don't love the way the R(t) looks...

PR #175, opened 2024-09-19
174 cmdstanr sample args

### LINKED ISSUE #174: Most if not all of the arguments to `cmdstanr` `sample()` should be optionally configurable from the `wwinference()` interface
Right now many are not. A power user might wish to configure them.

kaitejohnson:
Adding all non-deprecated arguments to `cmdstanr::sample()` to be configurable via `get_mcmc_options()`. Closes #174.

Also turn off messages for testing so that we are not getting CI warnings about ESS being capped. Note this doesn't seem to be working just by setting `show_messages = FALSE`, not seeing anything in global options either... 
https://mc-stan.org/cmdstanr/reference/cmdstanr_global_options.html

@athowes @seabbs Do you have any suggestions from your experience testing large stan models? Suppressing the output in CI can be a separate issue #176. Guess I can always just use `expect_warning()`


athowes:
What are the challenges here with testing it sorry? Some of the parameters have high Rhat because there are so many parameters?

dylanhmorris:
Need to discuss the way in which defaults are being specified here. I do not think expecting the user to call a function that just builds a list is ideal. 

Instead, I would suggest allowing the user just to provide the list directly. If you want to be able to fall back on defaults, I would do something like the following

```R
wwinference <- function(ww_data,
                        count_data,
                        forecast_date = NULL,
                        calibration_time = 90,
                        forecast_horizon = 28,
                        model_spec = get_model_spec(),
                        fit_opts = list(),
                        generate_initial_values = TRUE,
                        initial_values_seed = NULL,
                        compiled_model = compile_model()) {
   # ... other code here
    fit_opts_use <- get_default_fit_opts()
    fit_opts_use[names(fit_opts)] <- fit_opts  # this overwrites all and only the values the user sets in `fit_opts`
    # maybe check for invalid options here?
   # .. more code here
}
```

If you really want to enforce argument names you can check the final list names correspond to valid `$sample()` arguments

kaitejohnson:
> Need to discuss the way in which defaults are being specified here. I do not think expecting the user to call a function that just builds a list is ideal.
> 
> Instead, I would suggest allowing the user just to provide the list directly. If you want to be able to fall back on defaults, I would do something like the following
> 
> ```r
> wwinference <- function(ww_data,
>                         count_data,
>                         forecast_date = NULL,
>                         calibration_time = 90,
>                         forecast_horizon = 28,
>                         model_spec = get_model_spec(),
>                         fit_opts = list(),
>                         generate_initial_values = TRUE,
>                         initial_values_seed = NULL,
>                         compiled_model = compile_model()) {
>    # ... other code here
>     fit_opts_use <- get_default_fit_opts()
>     fit_opts_use[names(fit_opts)] <- fit_opts  # this overwrites all and only the values the user sets in `fit_opts`
>     # maybe check for invalid options here?
>    # .. more code here
> }
> ```
> 
> If you really want to enforce argument names you can check the final list names correspond to valid `$sample()` arguments

In this proposed syntax, what you are suggesting basically is not specifying the default to be in the `wwinference()` function, but instead within it (e.g. replace what you wrote as `get_default_fit_opts()` with the current `get_mcmc_options()`. 

This seems pretty similar to me, I don't think I am following how it is different other than the user calling a function to build a list vs it being done internally within the `wwinference()` function. Maybe missing something.

dylanhmorris:
As things are currently specified, the user is expected to configure options by doing:

```R
fit_opts = get_mcmc_options(# my choices of options as keyword arguments go here)
```
They _could_ also pass a list:

```
fit_opts = list(# my choices of options as list entries go here)
```
but then they wouldn't get the `wwinference` defaults for any unspecified options. Instead, `sample()` would fall back on its own defaults.

What I am suggesting is that a user shouldn't have to call a custom function of ours to override some but not all of our defaults, and should instead be able to pass a list.

kaitejohnson:
> As things are currently specified, the user is expected to configure options by doing:
> 
> ```r
> fit_opts = get_mcmc_options(# my choices of options go here, as keyword arguments
> )
> ```
> 
> They _could_ also pass a list:
> 
> ```r
> fit_opts = list(# my choices of options go here, as list entries
> )
> ```
> 
> but in that case they wouldn't get the `wwinference` defaults for any unspecified options. Instead, `sample()` would fall back on its own defaults.
> 
> What I am suggesting is that a user shouldn't have to call a custom function of ours to override some but not all of our defaults, and should instead be able to pass a list.

I see, and we'd make sure that the elements in that list are only the available args to `cmdstanr::sample()`?

dylanhmorris:
> I see, and we'd make sure that the elements in that list are only the available args to cmdstanr::sample()?

Yes, though it might also be reasonable to defer to `cmdstanr`'s own input validation there

tests/testthat/test_ww_model.R
```diff
@@ -2,12 +2,12 @@ test_that("Test the wastewater inference model on simulated data.", {
   #######
   # run model briefly on the simulated data
   #######
-  withr::with_seed(5, {
+  expect_warnin(withr::with_seed(5, {
```
dylanhmorris, line 5: Typo
```suggestion
  expect_warning(withr::with_seed(5, {
```

tests/testthat/test_wwinference.R
```diff
@@ -60,10 +60,15 @@ test_that("Function to get mcmc options produces the expected outputs", {
...
+    "chains", "seed", "adapt_delta", "max_treedepth",
+    "show_messages", "refresh", "save_latent_dynamics", "output_dir",
+    "output_basename", "sig_figs", "parallel_chains", "chain_ids",
+    "threads_per_chain", "opencl_ids", "save_warmup", "thin",
+    "adapt_engaged", "step_size", "metric", "metric_file", "inv_metric",
+    "init_buffer", "term_buffer", "window", "fixed_param", "show_exceptions",
+    "diagnostics", "save_metric", "save_cmdstan_config"
   )
-  # Checkmade doesn't work here for a list, says it must be a character vector
+  # Checkmare doesn't work here for a list, says it must be a character vector
```
dylanhmorris, line 71: Typo 
```suggestion
  # Checkmate doesn't work here for a list, says it must be a character vector
```

Also, this surprises me. The following code works for me:

```R
a_list <- list(
   a = 5,
   b = list(x=1, y=2),
   c = data.frame(x=1:5)
)
checkmate::check_names(names(a_list), must.include = c("a", "b", "c"))
```

kaitejohnson: Me too, that might be an old comment when I was using it incorrectly let me see

tests/testthat/test_wwinference.R
```diff
@@ -60,11 +60,15 @@ test_that("Function to get mcmc options produces the expected outputs", {
...
+    "show_messages", "refresh", "save_latent_dynamics", "output_dir",
+    "output_basename", "sig_figs", "parallel_chains", "chain_ids",
+    "threads_per_chain", "opencl_ids", "save_warmup", "thin",
+    "adapt_engaged", "step_size", "metric", "metric_file", "inv_metric",
+    "init_buffer", "term_buffer", "window", "fixed_param", "show_exceptions",
+    "diagnostics", "save_metric", "save_cmdstan_config"
   )
-  # Checkmade doesn't work here for a list, says it must be a character vector
-  expect_true(all(names(mcmc_options) %in% expected_names))
+  checkmate::check_names(names(mcmc_options), must.include = expected_names)
```
dylanhmorris, line 71:
```suggestion
  checkmate::expect_names(names(mcmc_options), must.include = expected_names)
```

dylanhmorris: note this is a stricter check than what you had before. Is it the one you want? Do you intend to have `get_mcmc_options()` define all those fields, or will it defer some to the `cmdstanr` defaults?

tests/testthat/test_wwinference.R
dylanhmorris, line 1: can use checkmate at l. 78 too:

```R
  expect_true(all(names(model_spec) %in% expected_names))
```
can be 

```
checkmate::expect_names(names(model_spec), must.include = expected_names)
```

tests/testthat/test_wwinference.R
dylanhmorris: Incidentally, for validation if you want to allow the user to set some but not all, but error if there are any unexpected arguments, you can use the `subset.of` argument to [`check/assert/expect_names()`](https://mllg.github.io/checkmate/reference/checkNames.html)

kaitejohnson: True, I think now that we are using `do.call()` the list doesn't explicitly pass the args so it just needs to be a subset of all possible cmdstanr args, though as written it will pass all of them.

I could add another test that ensure that the `fit_opts` specified by the user are a `subset_of` the cmdstanr sample args?

dylanhmorris: > True, I think now that we are using do.call() the list doesn't explicitly pass the args so it just needs to be a subset of all possible cmdstanr args, though as written it will pass all of them.

Correct, that's how it allows us more easily to defer some to the `cmdstanr` defaults, which was why I was suggesting above that `get_mcmc_options()` should maybe _not_ define all the fields, since there may be some where we'd prefer to defer to `cmdstanr` defaults.

> I could add another test that ensure that the fit_opts specified by the user are a subset_of the cmdstanr sample args?

Yes, that was what I was suggesting.

kaitejohnson: I wrote in the cmdstanr defaults (bc initially I thought that was needed to make it configurable). Happy to remove all the ones we added that are literally just the defaults repeated.

dylanhmorris: Yes, I would recommend that

R/wwinference.R
```diff
@@ -160,6 +157,12 @@ wwinference <- function(ww_data,
     )
   }
 
+  fit_opts_use <- get_mcmc_options() # get defaults
+  # this overwrites all and only the values the user sets in `fit_opts`
```
dylanhmorris, line 161:
```suggestion
  # this overwrites defaults with all and only the values the user sets in `fit_opts`
```

R/wwinference.R
```diff
@@ -31,13 +31,10 @@
...
-#' MCMC parameters as defined using `get_mcmc_options()`. This includes
-#' the following arguments, which are passed to
-#' [`$sample()`][cmdstanr::model-method-sample]:
-#' the number of chains, the number of warmup
-#' and sampling iterations, the maximum tree depth, the average acceptance
-#' probability, and the stan PRNG seed
+#' @param fit_opts The fit options to pass to
+#' [`$sample()`][cmdstanr::model-method-sample], must be a valid cmdstanr
+#' sample arg. Default is an empty list, which uses the
+#' default parameters in `get_mcmc_options()`
```
dylanhmorris, line 37:
```suggestion
#' @param fit_opts MCMC fitting options, as a list of keys and values. 
#' These are passed as keyword arguments to [`$sample()`][cmdstanr::model-method-sample].
#' Where no option is specified, [wwinference()] will fall back first on a package-specific default value 
#' defined by [get_mcmc_options()]. If no package-specific default exists, it will fall back on
#' the default value defined in [`$sample()`][cmdstanr::model-method-sample].
```

R/wwinference.R
```diff
@@ -138,7 +135,7 @@
 #'   fit_opts = get_mcmc_options(
 #'     iter_warmup = 250,
 #'     iter_sampling = 250,
-#'     n_chains = 2
+#'     chains = 2
```
dylanhmorris, line 138: This section needs revision given the changes to `fit_opts`

Good opportunity to review the whole example and try running it (since it isn't run for the docs due to the `\dontrun` command)

R/wwinference.R
```diff
@@ -160,6 +157,12 @@ wwinference <- function(ww_data,
     )
   }
 
+  fit_opts_use <- get_mcmc_options() # get defaults
+  # this overwrites all and only the values the user sets in `fit_opts`
+  fit_opts_use[names(fit_opts)] <- fit_opts
+
+  # maybe check for invalid options here?
```
dylanhmorris, line 164: can use `formalArguments(compiled_model$sample)` or `formals(compiled_model$sample)` to get arguments that  `compiled_model$sample()` expects (with `formalArguments`) or the argument _and_ their default values (with `formals(compiled_model$sample)`).

R/wwinference.R
```diff
@@ -375,17 +416,68 @@ fit_model <- function(compiled_model,
 get_mcmc_options <- function(
```
dylanhmorris, line 416: as discussed, remove causes where we use the `cmdstanr` defaults and include only cases where we overwrite it. 

Also need to consider whether you expect the user ever to call this function, or if it's just a way of documenting the defaults.

dylanhmorris (commented):
Some addition comments for the ongoing revisions

R/wwinference.R
kaitejohnson: I think I will change this to be an internal function

dylanhmorris: One problem I see with making it internal is that it makes the documentation less obvious. I think for now for simplicity of documentation I'd leave it exported unless you want to get rid of `get_mcmc_options()` entirely. In the latter case I'd suggest moving the small number of non-`cmdstanr-`default args configured in `get_mcmc_options()` to become top-level keyword arguments to `wwinference()` (i.e. you can do:

```R
wwinference(
    <arguments>,
    chains = 3)
```

dylanhmorris (commented):
A few comments / questions @kaitejohnson

R/wwinference.R
```diff
@@ -350,42 +369,49 @@ fit_model <- function(compiled_model,
...
     iter_warmup = 750,
     iter_sampling = 500,
-    n_chains = 4,
+    chains = 4,
+    parallel_chains = 4,
     seed = NULL,
     adapt_delta = 0.95,
-    max_treedepth = 12) {
+    max_treedepth = 12,
+    show_messages = TRUE) {
```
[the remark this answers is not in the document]
kaitejohnson: I think I had specified it because I thought this would be a useful and frequent one to modify, but since that is no longer the purpose of this function I will remove!

R/wwinference.R
```diff
@@ -350,42 +369,49 @@ fit_model <- function(compiled_model,
...
 #' arguments
-#' @export
 #'
-#' @examples
-#' mcmc_settings <- get_mcmc_options()
 get_mcmc_options <- function(
     iter_warmup = 750,
     iter_sampling = 500,
-    n_chains = 4,
+    chains = 4,
```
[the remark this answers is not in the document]
kaitejohnson: Same answer as above, will remove

R/wwinference.R
```diff
@@ -350,42 +369,49 @@ fit_model <- function(compiled_model,
...
+    parallel_chains = 4,
```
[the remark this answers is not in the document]
kaitejohnson: Can also remove this, but here the default is 1 (when I run `getOption("mc.cores", 1)` on my 8 core machine (I think) I get 1...)

kaitejohnson:
I'd prefer not to bloat the `wwinference()` call anymore, will just export the `get_mcmc_options()` function

R/wwinference.R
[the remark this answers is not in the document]
dylanhmorris: right, but you could set `options(mc.cores = 4)` or `options(mc.cores = 8)` on your machine if you wanted

kaitejohnson:
Oop the `silent_wwinference()` doesn't seem to be working properly

image

R/wwinference.R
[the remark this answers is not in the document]
kaitejohnson: Should we warn somewhere that the chains aren't by default running in parallel?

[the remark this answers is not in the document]
dylanhmorris: Yes, can just document it. Maybe demonstrate setting `options(mc.cores = <number>)` in the vignette.

R/wwinference.R
```diff
@@ -31,13 +31,13 @@
...
-#' the number of chains, the number of warmup
-#' and sampling iterations, the maximum tree depth, the average acceptance
-#' probability, and the stan PRNG seed
+#' @param fit_opts MCMC fitting options, as a list of keys and values.
+#' These are passed as keyword arguments to
+#' [`$sample()`][cmdstanr::model-method-sample].
+#' Where no option is specified, [wwinference()] will fall back first on a
+#' package-specific default value given by [get_mcmc_options()], if one exists.
+#' If no package-specific default exists, [wwinference()] will fall back on
+#' the default value defined in [`$sample()`][cmdstanr::model-method-sample].
```
dylanhmorris, line 40:
```suggestion
#' the default value defined in [`$sample()`][cmdstanr::model-method-sample].
#' See the documentation for [`$sample()`][cmdstanr::model-method-sample] for 
#' details on available options.
```

R/wwinference.R
```diff
@@ -31,13 +31,13 @@
...
-#' @param fit_opts The fit options, which in this case default to the
-#' MCMC parameters as defined using `get_mcmc_options()`. This includes
-#' the following arguments, which are passed to
-#' [`$sample()`][cmdstanr::model-method-sample]:
-#' the number of chains, the number of warmup
-#' and sampling iterations, the maximum tree depth, the average acceptance
-#' probability, and the stan PRNG seed
+#' @param fit_opts MCMC fitting options, as a list of keys and values.
+#' These are passed as keyword arguments to
+#' [`$sample()`][cmdstanr::model-method-sample].
```
dylanhmorris, line 36:
```suggestion
#' [`compiled_model$sample()`][cmdstanr::model-method-sample].
```

R/wwinference.R
```diff
@@ -160,6 +163,18 @@ wwinference <- function(ww_data,
...
   }
 
+  fit_opts_use <- get_mcmc_options() # get defaults
+  # this overwrites defaults with all and only the values the user sets in
+  # `fit_opts`
+  fit_opts_use[names(fit_opts)] <- fit_opts
+
+  # Check that the fit options passed to wwinference are valid cmdstanr::sample
+  # arguments
+  checkmate::assert_names(names(fit_opts),
```
dylanhmorris, line 173: add a test to `test_wwinference.R` to confirm this error gets raised when it should?

dylanhmorris (changes requested):
This is looking very good, thanks @kaitejohnson. A few final things, and then should be ready to go.

kaitejohnson:
Need to have chains even though its the default bc `wwinference()` has to know how many chains to run

R/wwinference.R
```diff
@@ -350,40 +370,41 @@ fit_model <- function(compiled_model,
 #' This function returns a list of MCMC settings to pass to the
 #' `cmdstanr::sample()` function to fit the model. The default settings are
```
dylanhmorris, line 371:
```suggestion
#' [`$sample()`][cmdstanr::model-method-sample] function to fit the model. The default settings are
```

R/wwinference.R
```diff
@@ -350,40 +370,41 @@ fit_model <- function(compiled_model,
...
 #' specified for production-level runs, consider adjusting to optimize
-#' for speed while iterating.
+#' for speed while iterating. All input arguments to `cmdstanr::sample()`
+#' are configurable by the user. See `cmdstanr::sample()` documentation
+#' https://mc-stan.org/cmdstanr/reference/model-method-sample.html for a full
+#' description of each argument.
```
dylanhmorris, line 376:
```suggestion
```suggestion
#' All input arguments to [`$sample()`][cmdstanr::model-method-sample]
#' are configurable by the user. See [`$sample()`][cmdstanr::model-method-sample] documentation
#' for details of the available arguments.
``````

File: R/wwinference.R [hunk 17]
```diff
@@ -350,40 +370,41 @@ fit_model <- function(compiled_model,
 #' This function returns a list of MCMC settings to pass to the
 #' `cmdstanr::sample()` function to fit the model. The default settings are
 #' specified for production-level runs, consider adjusting to optimize
```
dylanhmorris, line 372: I don't think the "consider adjusting for speed while iterating" is needed and it might be a bit confusing (since one of the configurable things is the number of iterations (in a different sense)

```suggestion
#' specified for production-level runs.
```

R/wwinference.R
```diff
@@ -350,40 +370,41 @@ fit_model <- function(compiled_model,
...
-#' default is `750`
+#' default is `750`.
 #' @param iter_sampling integer indicating the number of sampling iterations,
-#' default is `500`
-#' @param n_chains integer indicating the number of MCMC chains to run, default
-#' is `4`
-#' @param seed set of integers indicating the random seed of the stan sampler,
-#' default is NULL
+#' default is `500`.
+#' @param seed set of integers indicating the random seed of the Stan sampler,
```
dylanhmorris, line 383: Revise the description to be more consistent with [`sample()` docs](https://mc-stan.org/cmdstanr/reference/model-method-sample.html). NB:

> The only time seed should be specified as a vector (one element per chain) is if RNG functions are used in transformed data and the goal is to generate different data for each chain.

dylanhmorris (changes requested):
Thanks, @kaitejohnson. A few final things.

dylanhmorris (approved):
LGTM

PR #190, opened 2024-09-27
add multiple os to matrix strategy

kaitejohnson:
Attempting to test on ubuntu, windows, and mac...

### DIFF
--- a/.github/workflows/r-cmd-check.yaml
+++ b/.github/workflows/r-cmd-check.yaml
@@ -11,6 +11,7 @@ jobs:
       matrix:
         r-version: ["4.1.0", "release"]
         os: [windows-latest, ubuntu-latest]
+        rtools: ['44']
     runs-on:  ${{matrix.os}}
     steps:
     - uses: actions/checkout@v4
@@ -19,6 +20,7 @@ jobs:
         r-version: ${{matrix.r-version}}
         use-public-rspm: true
         extra-repositories: "https://mc-stan.org/r-packages/"
+        rtools-version: ${{ matrix.config.rtools }}
     - name: "Set up dependencies for wwinference"
       uses: r-lib/actions/setup-r-dependencies@v2
       with:
--- a/.github/workflows/r-cmd-check.yaml
+++ b/.github/workflows/r-cmd-check.yaml
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       matrix:
         r-version: ["4.1.0", "release"]
-        os: [windows-latest, ubuntu-latest]
+        os: [windows-latest, ubuntu-latest, macos-latest]
     runs-on:  ${{matrix.os}}
     steps:
     - uses: actions/checkout@v4
@@ -25,7 +25,7 @@ jobs:
       with:
         needs: check
     - name: "Install cmdstan via cmdstanr"
-      uses: epinowcast/actions/install-cmdstan@v1.5
+      uses: epinowcast/actions/install-cmdstan@v1
       with:
         cmdstan-version: "latest"
         num-cores: 2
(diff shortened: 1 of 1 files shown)


seabbs:
> Run Rscript -e 'cmdstanr::check_cmdstan_toolchain(fix = TRUE)'
> Installing mingw32-make with Rtools40
> The C++ toolchain required for CmdStan is setup properly!
> Run Rscript -e 'cmdstanr::install_cmdstan(version = "2.35.0", cores = 2)'
> Error: Error: unexpected numeric constant in "cmdstanr::install_cmdstan(version = 2.35.0"
> Execution halted
> Error: Process completed with exit code 1.

The remaining issue is with R 4.1 and is very odd. It looks to be due to the way that windows on that version of R only is interpreting strings?

seabbs:
Note this is currently pinned to a fixed release but needs to be changed back to the rolling release once that has been updated

seabbs:
@kaitejohnson I think cmdstanr wise this is now working and is "stable". The mac dependency installation issue is a odd one. I haven't seen the package that is causing the issue before.

.github/workflows/r-cmd-check.yaml
```diff
@@ -8,9 +8,11 @@ on:
 jobs:
   check-package:
     strategy:
+      fail-fast: false
```
seabbs, line 11: you might want to revert this as it makes the CI less wasteful (the whole matrix fails if one entry fails).

kaitejohnson: I probably copied this blindly from somewhere.

kaitejohnson:
Can you describe briefly what you changed here @seabbs ?

.github/workflows/r-cmd-check.yaml
seabbs: No I added it for testing so I could see what worked and what didn't.

seabbs: I prefer to have it on to avoid waste but you do get more information if its off.

seabbs:
I updated the rolling release of `install-cmdstan` to use the `gh` cli with a GITHUB secret rather than using the API directly with a API secret. I also switched to using javascript rather than custom bash and powershell across OS's to increase stability. 

The last change I needed to make in order to get this working with R 4.1 and windows (i.e it already worked on the latest release) was to be careful about the use of " and ' quotes. Apparently windows sometimes treated these interchangeable (but only when used in one order). 

Its not relevant here but I also added support for installing `cmdstanr` in the action if its not present in the calling action. I tried to do this in R but couldn't get it to only install R if not present vs all the time (mostly due actually neededing to try and detect the user having called `setup-r` in the wrapper step vs R (as the built in R can't be used to install packages).

The code changes here were almost entirely exploratory and should now be mostly set back to what you had. I found this still failed on mac but that this failure was unrelated to cmdstan (its something in your dependencies that I have never seen before).

kaitejohnson:
Wow thank you @seabbs. I'll pretend I understood 50% of that. 

I think I am in favor of the rolling release so we internally know when breaking changes are introduced (if I am understanding correctly), even 

Looks like the ruleset names got updated because these all ran.

dylanhmorris (approved):
This LGTM, and I will update the branch protection rules. Only one question: is there a reason not to add macOS now as well, @seabbs / @kaitejohnson?

seabbs:
Note also that all windows VM seem very slow at downloading cache objects which makes them run slower than they otherwise would. This is new behaviour as far as I am aware and so will hopefully be transient.

kaitejohnson:
@dylanhmorris I now can't seem to change the names of the checks anywhere -- I thought it was in the ruleset but it is not an option I am seeing...

dylanhmorris:
> @dylanhmorris I now can't seem to change the names of the checks anywhere -- I thought it was in the ruleset but it is not an option I am seeing...

I already did that, but for future reference:

`Settings > Rules > Rulesets > (click on the ruleset to modify)`

then scroll to "Branch rules"; find "Require status checks to pass", and click on "Show additional settings'

PR #191, opened 2024-09-27
Hot fix validate pmf

kaitejohnson:
This PR does two things, one which is necessary another which may be up for discussion but does appear to be necessary if we are passing in  long pmfs through a config... 
1. Fix the `validate_pmf()` function. `all.equal()` needs to be wrapped in `isTRUE()` because it returns either TRUE or a character with difference (alternatively we could use that error message here, but I like our custom one). https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/all.equal
2. Adds a test so that we catch something like this next time 

The up for debate thing is that:
1. Adjusting the tolerance. I loaded from the package data and passed to the config and I think passing things around leads to rounding error, so the long infection to hospital admissions delay has a difference of just above 1e-7. I think we should be a bit more flexible with this given we are setting this up for people to pass in delays via a config. 

Open to other suggestions e.g. for the pipeline we could one off load the rda from the `wwinference` package, use the package default, or compute directly but in general I think we want the `wwinference` package to allow users to specify via a config a long delay so tolerance can be a bit higher IMO.

### DIFF
--- a/R/validate.R
+++ b/R/validate.R
@@ -235,7 +235,7 @@ validate_pmf <- function(pmf,
                          count_data,
                          arg = "x",
                          call = rlang::caller_env()) {
-  if (!all.equal(sum(pmf), 1)) {
+  if (!isTRUE(all.equal(sum(pmf), 1))) {
     cli::cli_abort(
       c(
         "{.arg {arg}} does not sum to 1."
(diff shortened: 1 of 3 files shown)


dylanhmorris (dismissed):
LGTM. Maybe make the tolerance configurable?

dylanhmorris (approved):
LGTM

PR #193, opened 2024-09-30
Remove the snapshot test, fix a few other bugs

dylanhmorris:


tests/testthat/test_get_stan_data.R
```diff
@@ -199,9 +199,21 @@ test_that(paste0(
...
+
+  lab_site_subpop_spine_mod <- get_lab_site_subpop_spine(
+    lab_site_site_spine = lab_site_site_spine,
+    site_subpop_spine = site_subpop_spine_mod
+  )
+
   stan_data_ho <- get_stan_data(
     input_count_data,
-    input_ww_data = NULL,
+    input_ww_data = null_ww_data,
```
kaitejohnson, line 216: Why is this necessary? I think I wanted to actually test that passing in nothing for `input_ww_data` works as expected.

tests/testthat/test_ww_model.R
```diff
@@ -1,156 +0,0 @@
...
-      "eta_sd", "autoreg_rt", "log_r_mu_intercept", "sigma_rt",
-      "autoreg_rt_site", "i0_over_n", "sigma_i0", "sigma_growth",
-      "initial_growth", "inv_sqrt_phi_h", "sigma_ww_site_mean",
-      "sigma_ww_site_sd",
-      "p_hosp_w_sd", "t_peak", "dur_shed", "ww_site_mod_sd", "rt", "rt_site_t",
-      "p_hosp", "w", "hosp_wday_effect", "eta_i0", "eta_growth",
-      "infection_feedback", "p_hosp_mean"
-    )
-
-    for (param in model_params) {
```
kaitejohnson, line 81: I would advocate that we just delete this particular test, not this entire file. I think it is helpful to test that the model does run, the plots work etc.

kaitejohnson (changes requested):
Can you indicate what bugs were fixed here? I don't think we should remove the `test_ww_model.R` and `test_no_ww_model.R` files entirely, just the one comparing parameter results for each draw...

tests/testthat/test_get_stan_data.R
dylanhmorris: just because it's used in two places (also in the computation of `site_subpop_spine_mod`. If you want to test omitting the argument altogether I would delete the named argument rather than passing `NULL`.

tests/testthat/test_no_ww_model.R
```diff
@@ -1,49 +0,0 @@
-test_that("Test that the model runs on simulated data when include_ww=0.", {
```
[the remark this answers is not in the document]
dylanhmorris: As of a8dafa4 restored but pointed at the vignette data

kaitejohnson (approved):
Pending CI passes, this LGTM. Thank you for refactoring the test of the no wastewater model!

PR #198, opened 2024-10-01
Alternative install of cmdstan using cmdstanr::*

gvegayon:


seabbs:
Nice - this could be a great idea

gvegayon:
> Nice - this could be a great idea

The conda stuff did not work. Using the `cmdstanr::install_cmdstan()` works. It takes about 4 minutes, but I think it can be cached.

seabbs:
FYI that is exactly the same approach that the action is taking but without any bells and whistles

gvegayon:
This is now working thanks to @seabbs fix on the epinowcast action. Closing!

seabbs:
If you could detail somewhere why conda didn't work that would be great as it could be a neat solution someday

gvegayon:
For future reference, here is the commit with conda (which failed): <https://github.com/CDCgov/ww-inference-model/blob/d15fd25249004cd27dd761f43f2383dac14a7708/.github/workflows/r-cmd-check.yaml>

PR #199, opened 2024-10-01
hot-fix initialization misspecifies parameter name for `error_rt_subpop`

kaitejohnson:
Looks like I accidentally misnamed the initialization of the `error_rt_subpop` in `initialization.R` to `error_subpop`. I don't think our CI caught this bc it just prints as a stan warning, like the other stan warnings that we routinely ignore as the sampler is warming up :)

### DIFF
--- a/R/initialization.R
+++ b/R/initialization.R
@@ -105,7 +105,7 @@ get_inits_for_one_chain <- function(stan_data, stdev = 0.01) {
   )
 
   if (stan_data$n_subpops > 1) {
-    init_list$error_subpop <- matrix(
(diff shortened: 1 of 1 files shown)


seabbs:
great PR title

seabbs:
if this is about warnings from stan models vs anything else its because they print as output and aren't R warnings. You need to be using capture.output or similar to get them and then look at them

kaitejohnson:
Ah I see, yeah its a warning about the init that me missed. Will add to the description :)

dylanhmorris (approved):
LGTM

PR #201, opened 2024-10-02
Hot fix warn if data exceeds forecast date

kaitejohnson:
This came up while troubleshooting implementation with @hannahcohen4. 

The issue is right now it is on the user to make sure the data doesn't exceed the forecast date. Instead, I think we should still let them run the model, truncate the data, and give them a warning that we strongly recommend that vintaged datasets be used for retrospective forecasts. 

Curious others thoughts on this (it could also error). The point is just that as it is implemented this would only error in the `assert_dates_within_frame()`. We could modify this to be more verbose about which frame the data is not within (in this case, failed the check of being before than the forecast date).

### DIFF
--- a/R/checkers.R
+++ b/R/checkers.R
@@ -587,7 +587,8 @@ assert_dates_within_frame <- function(dates1,
   if (!check_dates2_win_frame) {
     cli::cli_abort(
       c(
-        "The two vectors of dates do not overlap",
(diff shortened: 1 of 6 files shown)


kaitejohnson:
Closing this in favor of more verbose error messaging rather than warnigns

PR #205, opened 2024-10-02
Update NEWS.md

kaitejohnson:
@gvegayon I think I want to leave the initial description of what the package does (maybe too verbose/redundant from README.md), but I have seen this style before e.g. I think EpiNow2/Epinowcast has this.


gvegayon (approved):
LGTM!

PR #208, opened 2024-10-02
Fix error messaging when data extends beyond forecast date

### LINKED ISSUE #202: Fix error messaging if accidentally pass data beyond forecast date
## Goal
Currently errors saying data is out of frame. We want to be more explicit, tell the user the data goes beyond the forecast date.

kaitejohnson:
This PR closes #202. It uses the `assert_no_dates_after_max()` check (that wasn't being used, since we removed needing to specify the forecast date from the preprocessing steps!). What it does:
- runs the assertion on the wastewater and count data within the `wwinference()` function
- updates the `assert_no_dates_after_max()` to take in args specifying what data is in the date vector and what the max date is
- updates the test of `assert_no_dates_after_max()` to expect the specific error message written

### DIFF
--- a/R/checkers.R
+++ b/R/checkers.R
@@ -588,7 +588,10 @@ assert_dates_within_frame <- function(dates1,
   checkmate::assert_date(dates1)
   checkmate::assert_date(dates2)
   check_dates2_win_frame <- min(dates1) <= max(dates2) &
-    min(dates2) >= min(dates1)
+    min(dates2) >= min(dates1) &
+    min(dates2) <= max(dates1) &
+    min(dates1) >= min(dates2)
+
   if (!check_dates2_win_frame) {
     cli::cli_abort(
       c(
--- a/R/checkers.R
+++ b/R/checkers.R
@@ -13,19 +13,26 @@
 #' @param date_vector vector of dates
 #' @param max_date string indicating the maximum date in ISO8601 convention
 #' e.g. YYYY-MM-DD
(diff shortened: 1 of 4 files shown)


R/checkers.R
```diff
@@ -13,19 +13,26 @@
...
     cli::cli_abort(
       c(
-        "The data passed in has observations beyond the specified",
-        "maximum date. Either this is the incorrect vintaged",
-        "data, or the data needs to be filtered to only contain",
-        "observations before the maximum date"
+        "The {.arg_dates {arg_dates}} passed in has observations beyond the",
+        " specified {.arg_max_date {arg_max_date}}. Either this is the ",
+        "incorrect vintaged data, or the data needs to be filtered to only ",
+        "contain observations before the {.arg_max_date {arg_max_date}}"
```
dylanhmorris, line 35:
```suggestion
        "The {.arg_dates {arg_dates}} passed in has observations after the ",
        "specified {.arg_max_date {arg_max_date}}. Check that this is the ",
        "dataset you intended to use with the given {.arg_max_date {arg_max_date}}."
```

dylanhmorris (changes requested):
Thanks @kaitejohnson. I think the error message should be simplified and generalized a bit while remaining informative. I think discussing vintaging and filtering is a bit too specific, given that there are other ways this check could fail.

dylanhmorris (dismissed):

dylanhmorris (approved):

PR #218, opened 2024-10-17
Issue 200: Modify plot methods

### LINKED ISSUE #200: Modify `plot` methods
## Goal
Realizing two issues with the current `plot()` method as we run on real data:
1. Lots of sites doesn't print well bc we are including the x axis dates in every facet. Might even want the option to turn off dates? 
2. The hospital admissions plots need evaluation data -- this should be optional since it won't be available in real-time!

kaitejohnson:
This PR closes #200. This makes a few minor changes:
- modifies plot of forecasted counts to not expect an evaluation dataset
- adds an example in the vignette to plot without the evaluation data
- replaces `scales = "free"` with `scales = "free_y"` so that the x-axis shared for the dates in the wastewater concentration plot. 

@gvegayon If you think there's a better way to handle the x axis dates (or that we should exclude them) lmk. I think best to have them as a default?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+#wwinference dev
+
+- Modify `plot_forecasted_counts()` so that it does not require an evaluation dataset ([#218](https://github.com/CDCgov/ww-inference-model/pull/218))
+
 # wwinference 0.1.0
 
 This is the first major release, focused on providing an initial version of the package.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-#wwinference dev
+# wwinference dev
 
 - Modify `plot_forecasted_counts()` so that it does not require an evaluation dataset ([#218](https://github.com/CDCgov/ww-inference-model/pull/218))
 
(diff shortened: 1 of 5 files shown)


NEWS.md
```diff
@@ -1,3 +1,7 @@
+#wwinference dev
```
dylanhmorris, line 1:
```suggestion
# wwinference dev
```

R/figures.R
dylanhmorris, line 1: Would be good to add a test that this can work both both `NULL` and non-`NULL` `count_data_eval` arguments.

Something as simple as confirming that ggplot plots are created without error in both cases with a tiny toy dataset.

dylanhmorris (changes requested):
Add a test and then this should be ready.

gvegayon:
I'm unsure about your question, @kaitejohnson, as I am no expert in ggplot2. All I can think is that I would (a) keep the detailing of the figures to a minimum or (b) ensure additional modifications are optional (e.g., `add_fancy_scales = TRUE`).

kaitejohnson:
Well it will default to plotting the dates so they all overlap and are unreadable, which is why I added the 45 degree angle dates... I think we should leave for now bc my understanding of ggplot is that the later thing takes precedent so the user can always override

kaitejohnson:
Need to add this to the change log (note to self)

dylanhmorris (approved):
LGTM, thanks @kaitejohnson

PR #219, opened 2024-10-17
Issue 197: rename `validate_both_datasets` function

kaitejohnson:


R/get_stan_data.R
```diff
@@ -433,7 +433,7 @@ get_stan_data <- function(input_count_data,
 
   # Validate both datasets if both are used----------------------------------
```
dylanhmorris, line 434:
```suggestion
# if both datasets are used, validate that that they are compatible and consistent with each other
```

dylanhmorris (commented):
One suggestion on a comment, which I'm just going to apply and merge

dylanhmorris (approved):

R/get_stan_data.R
```diff
@@ -431,9 +431,9 @@ get_stan_data <- function(input_count_data,
     arg_max_date = "forecast date"
   )
 
-  # Validate both datasets if both are used----------------------------------
+# if both datasets are used, validate that that they are compatible and consistent with each other
```
dylanhmorris, line 434:
```suggestion
# if both datasets are used, validate that that they are 
# compatible and consistent with each other
```

dylanhmorris (approved):

PR #224, opened 2024-10-18
Issue 215: Merge main to spatial

kaitejohnson:
Decided to work off of a separate branch off of spatial main with main merged in.

What this PR does:
- restructures the spatial model to also be relative to a reference subpopulation, which will always be the aux site. @cbernalz if you could review this and make sure it makes sense to you (I think there is redundancy in `sigma_rt` and the generalized variance, but I left both in since both were in the original implementation), that would be super helpful.
- modified the vignette to handle the new `get_draws()` function which no longer makes one df but a `wwinference_draws` object which contains multiple dfs. 

I believe this would now be in a place to be able to evaluate the forecast performance of wastewater concentrations fit to each model, since once #220 is merged (should be done shortly), that will be part of the package data (I will merge main into spatial main again once that gets merged). 

@arranhamlet Tagging for your awareness bc once this is merged you can use only spatial main for your analysis (instead of switching back and forth bc of the previous issue with the hospital admissions only model in `spatial-main`). Also happy for you to review these changes if you'd like.


kaitejohnson:
@cbernalz Just updated because the vignette had been using the deprecated `get_draws_df`, to now use `get_draws()` which returns a `wwinference_fit_draws` object now (basically a list with ech of the key variables as separate dataframes.

kaitejohnson:
Hmm still failing, will troubleshoot locally

cbernalz (approved):
This looks good!  Thanks!

kaitejohnson:
Ok @cbernalz so the prior on `sigma_generalized` that is being used in `spatial-main` is a lot more informative than the one we are using in `main` for `sigma_rt`.

In main we have:
`sigma_rt ~ N(0, 0.1)`
and in spatial main you have:
`sigma_generalized ~ logN( -5.3, 0.5)` which means the mean is 0.005 and the stdev is 0.003 in natural scale. 

I unfortunately can't remember the exact motivation for the choice of the prior here. My inclination is to try to parameterize both models the same way -- I am going to branch off of this and use `sigma_rt` as currently paramerized in `main` as `sigma_generalized` and look at the fits in the vignette as a test of the impact on. However, if you have (and if you do, apologies for missing it) a justification for the current values of this prior let me know.

kaitejohnson:
@cbernalz @arranhamlet Going to make the executive decision to merge this after our f2f discussion Friday acknowledging the following remaining issues:
- make a decision to consolidate `sigma_rt` from `main` and `sigma_generalized` from `spatial-main` #240 
- additional components of main will still need to be merged before results are "final", including what is already in main to add eval wastewater data to `generate_simulated_data()` function and the changes that are being worked on in the priors on infection feedback and the R(t) random walk step size.

PR #228, opened 2024-10-23
Modify infection feedback prior

kaitejohnson:
Originally was N(500,200) so lognormal(6, 0.4) which was a VERY strong infection feedback forcing term, especially at high incidence. 

This PR proposes changing this prior to be centered around 1 but in the range of o to 400 (exp(6) ~ 400). 

We plan to test this our in the `wweval` pipeline, look at performance, and then consider adding this to a version update.


seabbs:
Have you dug in to the original motivation for this? As this is likely to not be very informed by the length of data we are fitting to having a very weak prior means widening the prediction intervals essentially

kaitejohnson:
Closing this in favor of #236

PR #235, opened 2024-10-25
[Hot fix] test when pop size is not constant was failing

gvegayon:
Removed sampling of size pop and added an arbitrary site_pop value to ensure the pre_processing fails (passing the test 😋)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ## User-visible changes
 
-- `wwinference` now checks whether `site_pop` is unique per site (see issue [#223](https://github.com/CDCgov/ww-inference-model/issues/226) and reported by [@akeyel](https://github.com/akeyel)).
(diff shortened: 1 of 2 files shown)


kaitejohnson (dismissed):
LGTM, looks like CI isn't passing though?

gvegayon:
Yes, there is an error with rlang or something. @dylanhmorris, could you take a look too?

kaitejohnson:
@gvegayon Sorry for the delay on this, I found the issue which was that the `cbind` was making duplicate column names which was throwing off the error message. I made some changes but still used your suggested `seq_len()` to modify pop size.

kaitejohnson (dismissed):
LGTM now that is fixed!

gvegayon:
@dylanhmorris, can you approve this PR? We need another approve other than @kaitejohnson for merging.

dylanhmorris (approved):
One wording tweak you can take or leave. Thanks both for looking into this!

dylanhmorris (approved):

Issue #227, opened 2024-10-23
Infection feedback prior is too large


kaitejohnson:

dylanhmorris:
#236 will close this.

PR #236, opened 2024-10-30
Modify priors on `eta_sd` and `inf_feedback`

kaitejohnson:
This PR addresses #227. It modifies the stan model to take in a mean of the RW step size in R(t), `eta_sd`, and then it tests it using the subset of forecast dates and locations set up in https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/prod/_targets_subset_benchmarking.R

Benchmarking results can be found here in the files labeled as `subsets` https://github.com/CDCgov/wastewater-informed-covid-forecasting/tree/205-test-out-less-informative-eta-sd/output/benchmarking.  In the overall plots, the things to pay attention to are:
- pink (LHS) is this current proposal
- purple (RHS) is v1.0.0 (the currently, too strong infection feedback prior and two weak prior on `eta_sd`)

When we ran this on all forecast dates and location with a slightly more informative prior on `eta_sd`, the overall scores did look improved, though here the overall scores do not overall, if you look at the one by forecast date you see this is largely driven by 2024-01-08 as the pink (this proposed change) is an improvement most other times.

image


seabbs:
No object. Current proposal seems reasonable

dylanhmorris (dismissed):
LGTM, thanks @kaitejohnson

kaitejohnson:
Updated changelog

dylanhmorris (approved):

dylanhmorris (approved):

PR #239, opened 2024-10-31
Issue 238: Fix plot bug

kaitejohnson:
Currently the eval data gets added if present after the observation data and draws. The data the model is calibrated to is in the closed circles. and we want this to be in front of the white open circle eval data.

Note, I chose not to just cut off the eval data in the nowcast/forecast period, because there are changes in the values in the calibration period sometimes too, and we want these to be visible.

### DIFF
--- a/R/figures.R
+++ b/R/figures.R
@@ -57,7 +57,6 @@ get_plot_forecasted_counts <- function(draws,
       aes(x = .data$date, y = .data$pred_value, group = .data$draw),
       color = "red4", alpha = 0.1, linewidth = 0.2
     ) +
-    geom_point(aes(x = .data$date, y = .data$observed_value)) +
     geom_vline(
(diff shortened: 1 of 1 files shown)


gvegayon:
On the vignette, the only chance I see (compared to the website) is in this figure. New version:

plot-only-count-forecasts-1

Current vignette:

plot-only-count-forecasts-1-2

Was this expected?

kaitejohnson:
No! Good catch! Forgot to check since the preview doesn't work on my PC...

kaitejohnson:
This should be fixed now, was a typo. I just checked locally

kaitejohnson:
Confirmed both versions look as expected on the preview of the vignette on the site.

gvegayon (approved):

dylanhmorris:
Nice catch, thanks @gvegayon!

PR #240, opened 2024-11-02
Merge main to spatial sigma rt change

kaitejohnson:
This is a WIP/test run of fitting the model using the `sigma_rt` prior of `N(0,0.1)` that is currently being used in main, in the fits that incorporate spatial components.


kaitejohnson:
@cbernalz Looks like this is failing bc of the updates to `scoringutils` rather than any issues with the model. Going to band-aid fix by setting the version of scoringutils in the `DESCRIPTION` file for now, if not will make the changes

kaitejohnson:
But I am going to close this for now after realizing that the `sigma_rt` and `sigma_generalized` are not directly comparable

PR #255, opened 2024-11-24
Hot fix bug in ww data and ww data eval

kaitejohnson:
This PR fixes a bug in which the evaluation wastewater data and the calibration wastewater data generated in the `generate_simulated_data()` function were generated with independent sampling from daily data, resulting in non-identical datasets in the calibration period for the eval and calibration datasets.

This PR fixes this by by breaking the function into two, one which samples for frequency and on which samples for latency/truncates to create the calibration data

@cbernalz I checked and this should fix the date alignment problem! Sorry about that!


cbernalz (approved):
Looks good!

PR #257, opened 2024-11-25
Issue 256: Fix wastewater dataset alignment 

### LINKED ISSUE #256: Hot fix `ww_data ` and `ww_data_eval` not consistently sampled
As @cbernalz pointed out, the two wastewater data sets generated from `generate_simulated_data()` do not line up. This because I separately sample from the daily data twice to generate them in the current function. 

Solution:
- break into two functions, 1 which downsamples according to the prescribed frequencies to create the evaluation dataset, and a second which truncates after the last possible observed time and then applies the prescribed latency.

kaitejohnson:
This PR closes #256. It  breaks the previously used `downsample_ww_obs()` function into two functions, one which downsamples the entire time series, and another one which truncates the time series according to the prescribed latency at each lab-site pair and the known observation time and nowcast time.


cbernalz (approved):
Looks good!

Issue #259, opened 2024-12-08
Changing prior on exponential correlation function's $\phi$ parameter


cbernalz:
## Context

For the exponential correlation function in the spatial ww model, $\phi$ is centered at $0.25$.  This should be centered at $1$ because we already normalize the distance matrix by the largest distance.

## Requirements

- [x] Change `log_phi_mu_prior` and `log_phi_sd_prior` in `example_params.toml`.

PR #260, opened 2024-12-08
2024-12-08 update : changing center of prior.

cbernalz:
Regarding https://github.com/CDCgov/ww-inference-model/issues/259#issue-2725622751 .

### DIFF
--- a/inst/extdata/example_params.toml
+++ b/inst/extdata/example_params.toml
@@ -3,8 +3,8 @@ uot = 50
 
 [infection_process]
 # spatial params
-log_phi_mu_prior = -1.386294 # log(0.25)
-log_phi_sd_prior = 0.75
+log_phi_mu_prior = 0 # log(0.25)
+log_phi_sd_prior = 0.4
 l = 1
 log_sigma_generalized_mu_prior = -5.298317 # log(0.05^4)
 log_sigma_generalized_sd_prior = 0.5


kaitejohnson:
@cbernalz Could you provide a comparison of the posterior estimates of the phi parameter with both priors from the fit to the exponential and the data generated from the exponential? Ideally a density plot or histogram with the priors and posteriors overlaid

I am curious if with either of the prior values, the posterior estimate of the true phi is identifiable -- since this is generated with simulated data we know this. https://github.com/CDCgov/ww-inference-model/blob/bc89b7b4d76e0a6a862acbc25e44f5e5d147c923/vignettes/spatial_wwinference.Rmd#L246

After doing that check, I would also consider increasing phi in the generation of the simulated data above, this way the strength of the correlation as a function of distance is increased in your simulation analysis

cbernalz:
> @cbernalz Could you provide a comparison of the posterior estimates of the phi parameter with both priors from the fit to the exponential and the data generated from the exponential? Ideally a density plot or histogram with the priors and posteriors overlaid
> 
> I am curious if with either of the prior values, the posterior estimate of the true phi is identifiable -- since this is generated with simulated data we know this.
> 
> https://github.com/CDCgov/ww-inference-model/blob/bc89b7b4d76e0a6a862acbc25e44f5e5d147c923/vignettes/spatial_wwinference.Rmd#L246
> 
> After doing that check, I would also consider increasing phi in the generation of the simulated data above, this way the strength of the correlation as a function of distance is increased in your simulation analysis

@kaitejohnson Here is the plot that results from the prior and posterior for $\phi$.  The old prior is centered at 0.25 with sd 0.75.  The new prior is centered at 1 with sd 0.4.  The actual value is 1.5. 
IMG_7814

kaitejohnson:
This looks reasonable to me! Model seems to struggle in both cases a bit with identifiability but this makes sense -- between the R(t) variability and lab-site level observation noise there are a lot of drivers of variation in C(t).

kaitejohnson (approved):
LGTM!

PR #261, opened 2024-12-16
Issue 93:  Weekly count data feature

kaitejohnson:


seabbs:
This is a good example for the paper where you have missed out not being a true extension of EpiNow2 as a lot of work has been done on this to make it nice since you forked

PR #265, opened 2024-12-24
Reprex sample-based forecast to quantile-quantile coverage plot

kaitejohnson:
This is an example of how to convert a sample-based forecast to quantiles, and then to compute quantile-quantile coverage and create the plot we discussed f2f. 

This should be a fully reproducible example, and I am providing it just to make it really clear how to do this for the simulate data evaluation (for both wastewater concentrations and hospital admissions). 

Let me know if you have any issues @cbernalz and @arranhamlet, as we've also done this a few other ways outside of `scoringutils` that I can send along, but I do think this is the clearest and most straightforward!

### DIFF
--- a/scratch/scoringutils_example.R
+++ b/scratch/scoringutils_example.R
@@ -1,0 +1,36 @@
+library(scoringutils)
+# Using scoringutils 2.0.0
+
+# The sample data already has the prediction as `predictions`, the evaluation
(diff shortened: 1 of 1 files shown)


cbernalz (approved):
LGTM!  I realized what my mistake was, the sample_data was not in as_forecast_samples form.  This was extremely helpful, thank you!

